PR #9, opened 2021-06-22
[intrinsics] Intrinsics database and PDF generation for NEON and MVE.

fpetrogalli:


statham-arm:
In the intrinsics PDFs, perhaps it would be nicer to split up the "Argument preparation", "Instruction" and "Result" fields into multiple lines, by breaking the text up at the semicolons? That way the sequence of MVE assembly instructions would be written one per line, i.e. the usual way people like to read assembler. Also it would shorten the total line length, so that there'd be less awkward wrapping.

On the other hand, you'd probably end up with a much longer PDF by the 'number of pages' metric. But I think it would be a positive change overall.

fpetrogalli:
Hi @statham-arm , thank you for the review!

> In the intrinsics PDFs, perhaps it would be nicer to split up the "Argument preparation", "Instruction" and "Result" fields into multiple lines, by breaking the text up at the semicolons? That way the sequence of MVE assembly instructions would be written one per line, i.e. the usual way people like to read assembler. Also it would shorten the total line length, so that there'd be less awkward wrapping.
> 

Done. I have also split the intrinsic definition itself, and marked it as a code block so that it uses syntax highlighting. 

I am not happy with the fact that now the literal blocks are wrapped into a frame, but I haven't been able to find a way to configure `rst2pdf` so that the frame is removed...

> On the other hand, you'd probably end up with a much longer PDF by the 'number of pages' metric. But I think it would be a positive change overall.

Agree. The specs look much better now.

statham-arm:
Yes, that's much nicer. I hadn't even though of reformatting the intrinsic definition like that, and I agree that's an improvement as well!

tools/gen-intrinsics-specs.py
```diff
@@ -0,0 +1,718 @@
...
+
+def quote_split_intrinsics(intrinsic):
+    r"""
+    >>> quote_split_intrinsics('int f(int x, float y)')
+    '.. code:: c\n\n    int f(\n        int x,\n        float y)'
+
+    >>> quote_split_intrinsics('int f(int x)')
+    '.. code:: c\n\n    int f(int x)\n'
+    """
+    intrinsic_without_ending = intrinsic.removesuffix(')')
```
statham-arm, line 72: The `str.removesuffix` method doesn't exist before Python 3.9, which means this will fail on a lot of currently deployed Linux distros. Even the latest Ubuntu LTS (20.04) is still on 3.8.5.

I suggest either making this script compatible with older versions of Python, or if not, checking whether `sys.version_info >= (3,9)` at startup and generating a more friendly error message than `AttributeError: 'str' object has no attribute 'removesuffix'`.

tools/gen-intrinsics-specs.py
```diff
@@ -0,0 +1,718 @@
...
+                continue
+
+            print(
+                f"{classification_file}: skipping line {classification_db.line_num}: row = {row}", file=sys.stderr)
+
+    return classification_map
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="TODO")
```
statham-arm, line 697: Did you mean to fill in these TODOs, so that the `--help` output actually helps?

Perhaps a useful thing to put in the help for the three pathname options would be an example file in this repo that can be sensibly used with that option. That would be enough to let a user figure out a useful command line from scratch.

tools/gen-intrinsics-specs.py
fpetrogalli: I have chosen a different syntax, which should be compatible with many more versions of python.

tools/gen-intrinsics-specs.py
fpetrogalli: I have filled the TODO, but instead of adding a self explanatory example with CSV files I have added a self explanatory unit tests for `process_db`. Let me know if you think this id not as useful as you intended with the original suggestion.

tools/gen-intrinsics-specs.py
```diff
@@ -623,9 +626,78 @@ def get_section_data(row):
...
+    <BLANKLINE>
+    No category
+    ~~~~~~~~~~~
+    <BLANKLINE>
+    +-------------+-------+--------+---------+----------+
+    | T1          | T2    | T3     | T4      | T5       |
+    +=============+=======+========+=========+==========+
+    | .. code:: c | ::    | ::     | ::      | ``aaaa`` |
+    |             |       |        |         |          |
+    |     a A01() |     a |     aa |     aaa |          |
```
statham-arm, line 660: This test fails on Ubuntu 18.04. Instead of the nicely formatted table here, the output looks like this:
```
    +------+------+------+------+----------+
    | T1   | T2   | T3   | T4   | T5       |
    +======+======+======+======+==========+
    | .. code:: c
    <BLANKLINE>
        a A01()      | ::
    <BLANKLINE>
        a      | ::
    <BLANKLINE>
        aa      | ::
    <BLANKLINE>
        aaa      | ``aaaa`` |
    +------+------+------+------+----------+
```
It looks to me as if that's because 18.04's version of the `python3-tabulate` package isn't up-to-date enough to support multi-line records in tables.

It runs as expected on Ubuntu 20.04, though. (Between those releases, `python3-tabulate` went from 0.7.7 to 0.8.6.)

I suspect there's no good way to make that work, so perhaps checking the version of `tabulate` might be a simple answer, so that (again) we can generate a sensible error message?

fpetrogalli: I have added the check. Thank you for the review!

statham-arm (approved):
Thanks. LGTM now!

PR #17, opened 2021-07-19
[intrinsics] Use the `__ARM_NEON` macro for the Advanced SIMD support.

MattPD:
Fixes #15: Advanced SIMD support is available if the __ARM_NEON macro is predefined, https://developer.arm.com/documentation/101028/0012/13--Advanced-SIMD--Neon--intrinsics.

PR #19, opened 2021-07-22
acle: Correct usdot specification on AArch32

### LINKED ISSUE #18: usdot intrinsics missing AArch32 variants
None of the intrinsics for `usdot` should be `A64` only. They should all be `A64/A32`.

https://github.com/ARM-software/acle/blob/main/tools/intrinsic_db/advsimd.csv#L4258

The `(vector)` variant clearly has a `Q` variant, and for the `(by element)` intrinsics the higher lanes are implemented on AArch32 by
using the fact that the register file overlaps.

so if you want index [2,3] you remap the number and use the high half of the register.

e.g. a `Qn` consists of `Dn, Dn+1`. so when you want index e.g. 2, you remap `2->0` and use register `Dn+1` since the registers are always allocated in pairs.

TamarChristinaArm:
Fixes #18 by marking all the intrinsics as supported on AArch32.


fpetrogalli:
@TamarChristinaArm  - I think you are using an old version of the `next-release` branch. The new version has started a new changelog on top of the RST file (see #17). Please rebase on top of current next release and add an item to the changelog, somethign like `Correct usdot specification on AArch32`  (Note that you'll have to change also the template RST at `neon_intrinsics/advsimd.template.rst`.

Thanks!

TamarChristinaArm:
> Note that you'll have to change also the template RST at neon_intrinsics/advsimd.template.rst.

still have to do this.

fpetrogalli (approved):
Thanks @TamarChristinaArm !

PR #20, opened 2021-08-31
acle: Correct udot, sdot specification on AArch32

TamarChristinaArm:
Simular to the usdot intrinsics, these are all implementable on AArch32 using Dn/Dn+1 register pairs.


fpetrogalli (changes requested):
Only minor cosmetic changes.

Thank you!

TamarChristinaArm:
@fpetrogalli that should do it!

fpetrogalli (approved):

PR #21, opened 2021-09-13
Add Function Multi Versioning.

DanielKristofKiss:
Function Multi Versioning provides a convenient way to select the most
appropriate version of a function in runtime.
AArch32 features to be added later.

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -1978,42 +1978,43 @@ The following table lists the architectures feature mapping for AArch64
    | 50            | `FEAT_DotProd`           | dotprod       | ```ID_AA64ISAR0_EL1.DP == 0b0001```       |
    | 60            | `FEAT_SM3` && `FEAT_SM4` | sm            | ```ID_AA64ISAR0_EL1.SM4 == 0b0001 AND ``` <br> ```ID_AA64ISAR0_EL1.SM3 == 0b0001``` |
    | 70            | `FEAT_RDM`               | rdm           | ```ID_AA64ISAR0_EL1.RDM == 0b0001```      |
-   | 80            | `FEAT_LSE`               | lse           | ```ID_AA64ISAR0_EL1.LSE == 0b0001```      |
+   | 80            | `FEAT_LSE`               | lse           | ```ID_AA64ISAR0_EL1.Atomic == 0b0001```      |
    | 90            | Floating-point           | fp            | ```ID_AA64PFR0_EL1.FP != 0b1111```        |
    | 100           | Advanced SIMD            | advsimd       | ```ID_AA64PFR0_EL1.AdvSIMD != 0b1111```   |
    | 110           | crc32                    | crc32         | ```ID_AA64ISAR0_EL1.CRC32 == 0b0001```    |
    | 120           | `FEAT_SHA1`              | sha1          | ```ID_AA64ISAR0_EL1.SHA1 == 0b0001```     |
    | 130           | `FEAT_SHA256`            | sha256        | ```ID_AA64ISAR0_EL1.SHA2 == 0b0001```     |
    | 140           | `FEAT_SHA512`            | sha512        | ```ID_AA64ISAR0_EL1.SHA2 == 0b0010```     |
-   | 150           | `FEAT_AES`               | aes           | ```ID_AA64ISAR0_EL1.AES == 0b0001```      |
-   | 160           | `FEAT_FP16`              | fp16          | ```ID_AA64PFR0_EL1.FP == 0b0001```        |
-   | 170           | `FEAT_DIT`               | dit           | ```ID_AA64PFR0_EL1.DIT == 0b0001```       |
-   | 180           | `FEAT_DPB`               | dpb           | ```ID_AA64ISAR1_EL1.DPB >= 0b0001```      |
-   | 190           | `FEAT_DPB2`              | dpb2          | ```ID_AA64ISAR1_EL1.DPB  == 0b0002```     |
-   | 200           | `FEAT_JSCVT`             | jscvt         | ```ID_AA64ISAR1_EL1.JSCVT == 0b0001```    |
-   | 210           | `FEAT_FCMA`              | fcma          | ```ID_AA64ISAR1_EL1.FCMA == 0b0001```     |
-   | 220           | `FEAT_LRCPC`             | lrcpc         | ```ID_AA64ISAR1_EL1.LRCPC != 0b0000```    |
-   | 230           | `FEAT_LRCPC2`            | lrcpc2        | ```ID_AA64ISAR1_EL1.LRCPC == 0b0010```    |
-   | 240           | `FEAT_FRINTTS`           | frintts       | ```ID_AA64ISAR1_EL1.FRINTTS == 0b0001```  |
-   | 250           | SVE                      | sve           | ```ID_AA64PFR0_EL1.SVE != 0b0000 AND ``` <br> ```ID_AA64ZFR0_EL1.SVEver == 0b0000``` |
-   | 260           | `FEAT_BF16`              | sve_bf16      | ```ID_AA64ZFR0_EL1.BF16 != 0b0000```      |
-   | 270           | `FEAT_EBF16`             | sve_ebf16     | ```ID_AA64ZFR0_EL1.BF16 == 0b0010```      |
-   | 280           | `FEAT_I8MM`              | sve_i8mm      | ```ID_AA64ZFR0_EL1.I8MM == 0b00001```     |
-   | 290           | `FEAT_F32MM`             | sve_f32mm     | ```ID_AA64ZFR0_EL1.F32MM == 0b00001```    |
-   | 300           | `FEAT_F64MM`             | sve_f64mm     | ```ID_AA64ZFR0_EL1.F64MM == 0b00001```    |
-   | 310           | `FEAT_SVE2`              | sve2          | ```ID_AA64PFR0_EL1.SVE != 0b0000 AND ``` <br> ```ID_AA64ZFR0_EL1.SVEver == 0b0001``` |
-   | 320           | `FEAT_SVE_AES`           | sve_aes       | ```ID_AA64ZFR0_EL1.AES == 0b0001 OR ``` <br> ```ID_AA64ZFR0_EL1.AES == 0b0010``` |
-   | 330           | `FEAT_SVE_PMULL128`      | sve_pmull128  | ```ID_AA64ZFR0_EL1.AES == 0b0010```       |
-   | 340           | `FEAT_SVE_BitPerm`       | sve_bitperm   | ```ID_AA64ZFR0_EL1.BitPerm == 0b0001```   |
-   | 350           | `FEAT_SVE_SHA3`          | sve_sha3      | ```ID_AA64ZFR0_EL1.SHA3 == 0b0001```      |
-   | 360           | `FEAT_SME`               | sme           | ```ID_AA64PFR1_EL1.SME == 0b0001```       |
-   | 370           | `FEAT_MTE`               | mte           | ```ID_AA64PFR1_EL1.MTE >= 0b0001```       |
-   | 380           | `FEAT_MTE2`              | mte2          | ```ID_AA64PFR1_EL1.MTE >= 0b0010```       |
-   | 390           | `FEAT_MTE3`              | mte3          | ```ID_AA64PFR1_EL1.MTE >= 0b0011```       |
-   | 400           | `FEAT_SB`                | sb            | ```ID_AA64ISAR1_EL1.SB == 0b0001```       |
-   | 410           | `FEAT_SSBS`              | ssbs          | ```ID_AA64PFR1_EL1.SSBS == 0b0001```      |
-   | 420           | `FEAT_SSBS2`             | ssbs2         | ```ID_AA64PFR1_EL1.SSBS == 0b0010```      |
-   | 430           | `FEAT_BTI`               | bti           | ```ID_AA64PFR1_EL1.bt == 0b0010```        |
+   | 150           | `FEAT_AES`               | aes           | ```ID_AA64ISAR0_EL1.AES >= 0b0001```      |
+   | 160           | `FEAT_PMULL`             | pmull         | ```ID_AA64ISAR0_EL1.AES == 0b0010```      |
+   | 170           | `FEAT_FP16`              | fp16          | ```ID_AA64PFR0_EL1.FP == 0b0001```        |
+   | 180           | `FEAT_DIT`               | dit           | ```ID_AA64PFR0_EL1.DIT == 0b0001```       |
+   | 190           | `FEAT_DPB`               | dpb           | ```ID_AA64ISAR1_EL1.DPB >= 0b0001```      |
+   | 200           | `FEAT_DPB2`              | dpb2          | ```ID_AA64ISAR1_EL1.DPB  == 0b0002```     |
+   | 210           | `FEAT_JSCVT`             | jscvt         | ```ID_AA64ISAR1_EL1.JSCVT == 0b0001```    |
+   | 220           | `FEAT_FCMA`              | fcma          | ```ID_AA64ISAR1_EL1.FCMA == 0b0001```     |
+   | 230           | `FEAT_LRCPC`             | lrcpc         | ```ID_AA64ISAR1_EL1.LRCPC != 0b0000```    |
+   | 240           | `FEAT_LRCPC2`            | lrcpc2        | ```ID_AA64ISAR1_EL1.LRCPC == 0b0010```    |
+   | 250           | `FEAT_FRINTTS`           | frintts       | ```ID_AA64ISAR1_EL1.FRINTTS == 0b0001```  |
+   | 260           | SVE                      | sve           | ```ID_AA64PFR0_EL1.SVE != 0b0000 AND ``` <br> ```ID_AA64ZFR0_EL1.SVEver == 0b0000``` |
+   | 270           | `FEAT_BF16`              | sve_bf16      | ```ID_AA64ZFR0_EL1.BF16 != 0b0000```      |
+   | 280           | `FEAT_EBF16`             | sve_ebf16     | ```ID_AA64ZFR0_EL1.BF16 == 0b0010```      |
+   | 290           | `FEAT_I8MM`              | sve_i8mm      | ```ID_AA64ZFR0_EL1.I8MM == 0b00001```     |
+   | 300           | `FEAT_F32MM`             | sve_f32mm     | ```ID_AA64ZFR0_EL1.F32MM == 0b00001```    |
+   | 310           | `FEAT_F64MM`             | sve_f64mm     | ```ID_AA64ZFR0_EL1.F64MM == 0b00001```    |
+   | 320           | `FEAT_SVE2`              | sve2          | ```ID_AA64PFR0_EL1.SVE != 0b0000 AND ``` <br> ```ID_AA64ZFR0_EL1.SVEver == 0b0001``` |
+   | 330           | `FEAT_SVE_AES`           | sve_aes       | ```ID_AA64ZFR0_EL1.AES == 0b0001 OR ``` <br> ```ID_AA64ZFR0_EL1.AES == 0b0010``` |
+   | 340           | `FEAT_SVE_PMULL128`      | sve_pmull128  | ```ID_AA64ZFR0_EL1.AES == 0b0010```       |
+   | 350           | `FEAT_SVE_BitPerm`       | sve_bitperm   | ```ID_AA64ZFR0_EL1.BitPerm == 0b0001```   |
+   | 360           | `FEAT_SVE_SHA3`          | sve_sha3      | ```ID_AA64ZFR0_EL1.SHA3 == 0b0001```      |
+   | 370           | `FEAT_SME`               | sme           | ```ID_AA64PFR1_EL1.SME == 0b0001```       |
+   | 380           | `FEAT_MTE`               | mte           | ```ID_AA64PFR1_EL1.MTE >= 0b0001```       |
+   | 390           | `FEAT_MTE2`              | mte2          | ```ID_AA64PFR1_EL1.MTE >= 0b0010```       |
+   | 400           | `FEAT_MTE3`              | mte3          | ```ID_AA64PFR1_EL1.MTE >= 0b0011```       |
+   | 410           | `FEAT_SB`                | sb            | ```ID_AA64ISAR1_EL1.SB == 0b0001```       |
+   | 420           | `FEAT_SSBS`              | ssbs          | ```ID_AA64PFR1_EL1.SSBS == 0b0001```      |
+   | 430           | `FEAT_SSBS2`             | ssbs2         | ```ID_AA64PFR1_EL1.SSBS == 0b0010```      |
+   | 440           | `FEAT_BTI`               | bti           | ```ID_AA64PFR1_EL1.bt == 0b0010```        |
 
 ### Selection
 
--- a/main/acle.md
+++ b/main/acle.md
@@ -271,6 +271,7 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * Moved the [Future directions](#future-directions) chapter to the end.
 * Added a description of support levels in [Current Status and
   Anticipated Changes](#current-status-and-anticipated-changes).
+* Support added for [Function Multi Versioning](#function-multi-versioning).
 
 ### References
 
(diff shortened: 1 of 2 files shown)


main/acle.rst
```diff
@@ -2226,6 +2226,201 @@ or compiler-generated functions.
...
+   | 60           | ``FEAT_SM3`` && ``FEAT_SM4``        | sm                  | ID_AA64ISAR0_EL1.SM4 == 0b0001 &&   |
+   |              |                                     |                     | ID_AA64ISAR0_EL1.SM3 == 0b0001      |
+   +--------------+-------------------------------------+---------------------+-------------------------------------+
+   | 70           | ``FEAT_RDM``                        | rdm                 | ID_AA64ISAR0_EL1.RDM == 0b0001      |
+   +--------------+-------------------------------------+---------------------+-------------------------------------+
+   | 80           | ``FEAT_LSE``                        | lse                 | ID_AA64ISAR0_EL1.LSE == 0b0001      |
+   +--------------+-------------------------------------+---------------------+-------------------------------------+
+   | 90           | crc32                               | crc32               | ID_AA64ISAR0_EL1.CRC32 == 0b0001    |
+   +--------------+-------------------------------------+---------------------+-------------------------------------+
+   | 100          | ``FEAT_SHA1``                       | sha1                | ID_AA64ISAR0_EL1.LSE == 0b0001      |
```
enh-google, line 2331: copy & paste in the last column.

enh-google (commented):
LSE's one of the ones i happen to be interested in :-)

nsz-arm (commented):
if you plan to use ifuncs for this then keep in mind that only HWCAP based selection is simple (that is passed to resolvers), anything else requires per module cache and per module feature query that gives constant result during the lifetime of a process (yes x86 queries cpuid per module and caches the results per module).

main/acle.rst
```diff
@@ -2226,6 +2226,201 @@ or compiler-generated functions.
...
+----------
+
+The following rules shall be followed by all implementations.
+Implementation of the selection algoritm is platform dependent.
+
+Only those versions could be considered where all dependent features are available.
+
+Rules of version selection are in order:
+
+#. If possible select the CPU implementation specific version else
```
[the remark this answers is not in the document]
DanielKristofKiss: Agree, in case of bigLITTLE the cpu implementation selection is complicated so I add a note about it. I don't think is will be widely used but I would keep it in the spec. On a given platforms it might be disabled in compile time.

ifunc resolver could be used I think but the resolver might use the  cpu feature register emulation.
https://www.kernel.org/doc/html/latest/arm64/cpu-feature-registers.html

> i think the specification does not make it clear that for the lifetime of a given process only one variant is selected.
Good catch, I make it clear.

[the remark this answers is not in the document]
nsz-arm: cpu feature register emulation does not solve all problems: ifuncs are problematic if there are any selection rules that may change value at runtime. in this case MIDR is different across cores so if that is used for selection then you have to ensure it is only read once and that result is used consistently throughout the process, which i dont know how to do with ifuncs. (reading MIDR multiple times can give different results if the process gets scheduled to different cores).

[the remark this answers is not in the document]
enh-google: yeah, non-homogeneous SoCs are a PITA. for Android's libc with arm64 the only variant we have right now is MTE, so we just use hwcap there and assume that no-one would ship a chip where only some cores support it: https://android.googlesource.com/platform/bionic/+/refs/heads/master/libc/arch-arm64/dynamic_function_dispatch.cpp

but that might be more problematic for less fundamental features.

/me wonders whether we should change the CDD to outlaw SoCs where hwcap doesn't report the same value for every core?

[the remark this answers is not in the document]
DanielKristofKiss: Kernel ensures the provided HWCAPs\CPU Feature Registers are consistent across the cores and supported by the current kernel.
The only exception is MIDR in the CPU Feature Registers.
Seems out there only a few cases that really cares about the actual CPU so maybe dropping this option won't cause any trouble later.

main/acle.rst
```diff
@@ -2226,6 +2226,201 @@ or compiler-generated functions.
...
+   +--------------+-------------------------------------+---------------------+-------------------------------------+
+   | 110          | ``FEAT_SHA256``                     | sha256              | ID_AA64ISAR0_EL1.SHA2 == 0b0001     |
+   +--------------+-------------------------------------+---------------------+-------------------------------------+
+   | 120          | ``FEAT_SHA512``                     | sha512              | ID_AA64ISAR0_EL1.SHA2 == 0b0010     |
+   +--------------+-------------------------------------+---------------------+-------------------------------------+
+   | 130          | ``FEAT_AES``                        | aes                 | ID_AA64ISAR0_EL1.AES == 0b0001      |
+   +--------------+-------------------------------------+---------------------+-------------------------------------+
+   | 140          | Floating-point                      | fp                  | ID_AA64PFR0_EL1.FP != 0b1111        |
+   +--------------+-------------------------------------+---------------------+-------------------------------------+
+   | 150          | Advanced SIMD                       | advsimd             | ID_AA64PFR0_EL1.AdvSIMD != 0b1111   |
```
fpetrogalli, line 2341: Do we need `advsimd` and `sve` to be somehow aware of / compatible with the mangling of the names of the Vector Function ABI mangling? https://github.com/ARM-software/abi-aa/blob/main/vfabia64/vfabia64.rst#vector-function-name-mangling

Also, which of the two available calling conventions for NEON is the `advsimd` version supposed to follow? The standard CC or the `aarch64_vector_pcs` one? The latter is more efficient for leaf functions that are passing values via NEON registers. (see https://github.com/ARM-software/abi-aa/blob/main/vfabia64/vfabia64.rst#vector-procedure-call-standard) The

fpetrogalli (changes requested):
Hi @DanielKristofKiss - thank you for submitting this.

I have added few notes.

Could you please start a section after the section `Changes between ACLE Q2 2017 and ACLE Q2 2018` on line 265, `Changes for next release`, and add a sentence/item about the function multi-versioning support?

Thank you!

main/acle.rst
DanielKristofKiss: I'll add a note, FMV won't change the calling convention because the caller always will call the function in the same way.

fpetrogalli: I have two cases in my mind.

1. What happens if someone defines a NEON function with `aarch64_vector_pcs` and the compiler decides to use multi-versioning on it?
2. What happens if the compiler does multi-versioning of a function that does not use NEON registers in the parameter list, but that gets auto-vectorized to execute in the NEON units?

DanielKristofKiss: 1. The function declaration should have the `__attribute__ ((aarch64_vector_pcs))` and all function definition too. If only one of the versions would uses the vertor pcs then it won't work since the caller has no information about which version will run.  This example should work:
``` 
void __attribute__ ((aarch64_vector_pcs)) foo(uint32x2_t a);
//...
void __attribute__ ((aarch64_vector_pcs ,target_varinant("default")) foo(uint32x2_t a){}
void __attribute__ ((aarch64_vector_pcs ,target_varinant("advsimd")) foo(uint32x2_t a){}
```

2. In this case the function should just should conform to the aapcs as it would do without multi-versioning.
Following example generates two `foo` functions where the sve2 variant may be autovectorised.
```
void __attribute__ ((target_clones("sve2")) foo(){}
```

nsz-arm:
i think it is not clear from this spec how existing acle features may be used in a target_version function: so far acle features were per translation unit and selected based on cflags. e.g. someone might expect mte acle syntax and intrinsics are available in a target_version("mte") function, while currently that's only guaranteed when __ARM_FEATURE_MEMORY_TAGGING is defined for the entire tu. i guess the safest is to say that c language level extensions are still per tu, only the code generation is affected by the target_version not the c syntax.

fpetrogalli:
@DanielKristofKiss -  kindly rebase your code against origin/next-release. I have moved the tip of next-release to the current release (2021Q3)

main/design_documents/function-multi-versioning.rst
```diff
@@ -0,0 +1,56 @@
...
+Attributes
+##########
+
+These use cases to be considered:
+- The compiler is not aware of the new attributes.
+Usually unknown attributes are ignored, but the feature test macro is there to
+disable the unwanted code parts.
+- FMV aware compiler processes the attributes with newer arch features.
+In this case the compiler should ignore the new feature because it won't able to
+process it.
```
fpetrogalli, line 53: This doesn't render correctly in the html page: https://github.com/DanielKristofKiss/acle/blob/FMV/main/design_documents/function-multi-versioning.rst#attributes

fpetrogalli (changes requested):
Hi @DanielKristofKiss  - I have added some comments.

Kind regards,

Francesco

main/acle.rst
```diff
@@ -2242,6 +2246,204 @@ or compiler-generated functions.
...
+   | 420            | ``FEAT_SSBS2``                      | ssbs2               | ID_AA64PFR1_EL1.SSBS == 0b0010      |
+   +----------------+-------------------------------------+---------------------+-------------------------------------+
+   | 430            | ``FEAT_BTI``                        | bti                 | ID_AA64PFR1_EL1.bt == 0b0010        |
+   +----------------+-------------------------------------+---------------------+-------------------------------------+
+
+Selection
+---------
+
+The following rules shall be followed by all implementations. Implementation of
+the selection algorithm is platform dependent. The selection is permanent for
```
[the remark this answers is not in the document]
DanielKristofKiss: Platform means the CPU/Vendor/OS as in the target triplet.
The user's code will be shared across all platforms so I would avoid to have different behaviour on different platforms so one algorithm, set of rules for implementation hopefully enough.

main/acle.rst
```diff
@@ -2242,6 +2246,204 @@ or compiler-generated functions.
...
+the lifetime of the process. Only those versions could be considered where all
+dependent features are available.
+
+Rules of version selection are in order:
+
+1. Select the most specific version else
+
+2. select the version with the highest priority else
```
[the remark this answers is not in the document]
DanielKristofKiss: Priorities are defined in the tables above.

main/design_documents/function-multi-versioning.rst
```diff
@@ -0,0 +1,56 @@
...
+user space.
+
+Why ACLE instead of ABI? FMV shall not change how the function behaves from the
+ABI perspective. All versions shall have the same function signature, because
+the caller will not know which version is actually called.
+
+Today the ``target`` attribute is used for many purposes and once code will
+start use function multi versioning the not yet updated compilers may got
+confused by the new values therefore I propose the ``target_version`` besides
+the ``target_clones``.
```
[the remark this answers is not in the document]
DanielKristofKiss: done,
FMV on other arch uses the target and target_clone, but we add target_version and target_clones to support the same.

main/design_documents/function-multi-versioning.rst
```diff
@@ -0,0 +1,56 @@
...
+I would like to propose the addition of the following feature test macro, to be
+defined to 1 if the support is available:
+
+::
+
+  __ARM_FEATURE_FUNCTION_MULTI_VERSIONING
+
+I don't expect that enabling a feature for function then ACLE features will be
+available in the function scope if otherwise won't be enabled. This would require
+significant changes how the C/C++ preprocessor works.
```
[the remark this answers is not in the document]
DanielKristofKiss: fixed/

main/acle.rst
```diff
@@ -2242,6 +2246,204 @@ or compiler-generated functions.
 
 This attribute does not apply to AArch64.
 
+Function Multi Versioning (FMV)
+===============================
+
+Function Multi Versioning provides a convenient way to select the most
+appropriate version of a function in runtime. All versions of the function may
```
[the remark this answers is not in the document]
DanielKristofKiss: done

main/design_documents/function-multi-versioning.rst
DanielKristofKiss: fixed.

main/design_documents/function-multi-versioning.rst
```diff
@@ -0,0 +1,56 @@
...
+form a desktop operating system. On Linux based systems I expect it will use
+IFUNC[2]. While the ACLE should specify the behavior that software could rely on
+every implementation.
+
+Feature test macro
+##################
+
+In order for users to determine the availability of support in the toolchain,
+I would like to propose the addition of the following feature test macro, to be
+defined to 1 if the support is available:
```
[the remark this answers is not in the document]
DanielKristofKiss: done

main/design_documents/function-multi-versioning.rst
```diff
@@ -0,0 +1,56 @@
...
+should use the same source code formats because without it the software
+development and deployment will be too complicated and the feature won't be
+used. Arm architectures are a bit more complex in view of feature availability
+therefore function versions would be better to be bound to features instead of
+cpu implementations. Also the kernels may limit the available features in the
+user space.
+
+Why ACLE instead of ABI? FMV shall not change how the function behaves from the
+ABI perspective. All versions shall have the same function signature, because
+the caller will not know which version is actually called.
```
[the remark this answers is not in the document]
DanielKristofKiss: done

main/design_documents/function-multi-versioning.rst
```diff
@@ -0,0 +1,56 @@
+=======================================================
+Function Multi Versioning support for Arm architectures
+=======================================================
+
+I'd like to propose this addition to the ACLE. Other architectures supports
```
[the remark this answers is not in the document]
DanielKristofKiss: done, thanks!

main/design_documents/function-multi-versioning.rst
```diff
@@ -0,0 +1,56 @@
...
+
+Today the ``target`` attribute is used for many purposes and once code will
+start use function multi versioning the not yet updated compilers may got
+confused by the new values therefore I propose the ``target_version`` besides
+the ``target_clones``.
+
+The ACLE won't specify the underlying mechanism of the selection and the
+dispatching. On an embedded system probably it will be implemented differently
+form a desktop operating system. On Linux based systems I expect it will use
+IFUNC[2]. While the ACLE should specify the behavior that software could rely on
```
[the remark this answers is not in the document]
DanielKristofKiss: done

main/acle.rst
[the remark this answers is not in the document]
DanielKristofKiss: done

main/design_documents/function-multi-versioning.rst
```diff
@@ -0,0 +1,56 @@
+=======================================================
+Function Multi Versioning support for Arm architectures
+=======================================================
+
+I'd like to propose this addition to the ACLE. Other architectures supports
+function multi versioning[1]. In sort just FMV. All toolchains and platforms
```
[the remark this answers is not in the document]
DanielKristofKiss: done

main/acle.rst
```diff
@@ -273,6 +273,10 @@ Changes between ACLE Q2 2021 and ACLE Q3 2021
 * Introduce ``__ARM_FEATURE_PAUTH`` and ``__ARM_FEATURE_BTI`` in sections
   ssec-PAC_ and ssec-BTI_ respectively.
 
+Changes for the next release
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* Support added for Function Multi Versioning.
```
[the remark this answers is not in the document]
DanielKristofKiss: done

main/acle.rst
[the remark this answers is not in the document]
fpetrogalli: Would it make sense then to change it to this?

```suggestion
the selection algorithm is platform dependent, where with platform we mean CPU/Vendor/OS as in the target triplet. The selection is permanent for
```

main/acle.rst
```diff
@@ -2255,6 +2257,206 @@ or compiler-generated functions.
...
+   | 350            | ``FEAT_SVE_SHA3``                   | sve_sha3            | ID_AA64ZFR0_EL1.SHA3 == 0b0001      |
+   +----------------+-------------------------------------+---------------------+-------------------------------------+
+   | 360            | ``FEAT_SME``                        | sme                 | ID_AA64PFR1_EL1.SME == 0b0001       |
+   +----------------+-------------------------------------+---------------------+-------------------------------------+
+   | 370            | ``FEAT_MTE``                        | mte                 | ID_AA64PFR1_EL1.MTE >= 0b0001       |
+   +----------------+-------------------------------------+---------------------+-------------------------------------+
+   | 380            | ``FEAT_MTE2``                       | mte2                | ID_AA64PFR1_EL1.MTE >= 0b0010       |
+   +----------------+-------------------------------------+---------------------+-------------------------------------+
+   | 390            | ``FEAT_MTE3``                       | mte3                | ID_AA64PFR1_EL1.MTE >= 0b0010       |
```
DanielKristofKiss, line 2433: typo `ID_AA64PFR1_EL1.MTE >= 0b0011`

DanielKristofKiss: fixed

DanielKristofKiss:
Change is rebased, now in markdown format. Typos are fixed, suggestions are added.

DanielKristofKiss:
sorry for the noise, fix the rendering of the table on Github.

main/acle.md
```diff
@@ -1854,6 +1856,142 @@ or compiler-generated functions.
 
 This attribute does not apply to AArch64.
 
+## Function Multi Versioning
+
+The specification for Function Multi Versioning is in `BETA` state and may
+change or be extended in the future.
```
fpetrogalli, line 1862:
```suggestion
The specification for Function Multi Versioning is in [**Beta** state](#current-status-and-anticipated-changes)
and may change or be extended in the future.
```

main/acle.md
```diff
@@ -1854,6 +1856,142 @@ or compiler-generated functions.
...
+
+Function Multi Versioning provides a convenient way to select the most
+appropriate version of a function at runtime. All versions of the function may
+be in the final binary. The compiler generates all supported versions and the
+runtime makes the selection at load time.
+
+The following attributes trigger the multi variant code generation:
+`target_version`, `target_clones`. These attributes can't be mixed. The
+"default" variant means the version of the function that would be generated
```
fpetrogalli, line 1871: Should this be the following? As far as I understand `"default"` and not just `default` needs to be passed to the attribute.

```suggestion
`"default"` variant means the version of the function that would be generated
```

main/acle.md
```diff
@@ -1854,6 +1856,142 @@ or compiler-generated functions.
...
+without these attributes.
+
+This feature may be turned off in compile time. In this case the "default"
```
fpetrogalli, line 1874:
```suggestion
This feature may be disabled at compile time. In this case the `"default"`
```

main/acle.md
```diff
@@ -1854,6 +1856,142 @@ or compiler-generated functions.
...
+variant shall be used.
+
+`__ARM_FEATURE_FUNCTION_MULTI_VERSIONING` is defined to 1 if available.
```
fpetrogalli, line 1877:
```suggestion
`__ARM_FEATURE_FUNCTION_MULTI_VERSIONING` is defined to 1 if the versioning mechanism described in this section is available in the compiler.
```

main/acle.md
```diff
@@ -1854,6 +1856,142 @@ or compiler-generated functions.
...
+
+``` c
+    __attribute__((target_version("name")))
+```
+
+when applied to a function it will become one of the versions. Function with the
+same name may exist with multiple versions in the same translation unit. One
+"default" version of the  function is required to be provided. Implicitly, when
+without attribute, or explicitly providing the "default" in the attribute. If
+only the "default" variant exist it should be linked directly.
```
fpetrogalli, line 1887: Can we make these an item list? Something like:

```
The attribute `__attribute__((target_version("name")))` expresses the following:

1. when applied to a function it will become one of the versions. Function with the
same name may exist with multiple versions in the same translation unit.
2. One
"default" version of the  function is required to be provided.
3. Implicitly, when
without attribute, or explicitly providing the "default" in the attribute.
4. If
only the "default" variant exist it should be linked directly.
```

Same for the paragraph below about `target_clones`.

In general I prefer to explicitly enumerate/itemize the rules so that they are easier to refer to.

main/acle.md
```diff
@@ -1854,6 +1856,142 @@ or compiler-generated functions.
...
+
+``` c 
+    __attribute__((target_clones("name",...)))
+```
+
+when applied to a function the compiler emits multiple versions based on the
+arguments. One of them is implicitly the "default". If a name is not recognized
+the compiler should ignore it. If the default matches with an other explicitly
+provided version the compiler can emit only one function instead of the two.
```
fpetrogalli, line 1896: same here, use number lists

main/acle.md
```diff
@@ -1854,6 +1856,142 @@ or compiler-generated functions.
...
+
+Dependent features names could be joined by the `+` sign.
+
+None of these attributes will enable the corresponding ACLE features.
+
+These attributes have no effect on the calling convention, all versions must
+use the same calling convertion.
```
fpetrogalli, line 1903: Given that these sentences are additional rules, and that the top of the paragraph have also rules, I would organize this paragraph as follows:

1. Introduce the concept of function multiversioning
2. introduce the names of the 2 attributes
3. introduce the rules that are common to both attributes, with enumerated list
4. introduce the rules that are specific to each attributes, with enumeration lists that are starting from the previous set of enumeration list. By doing so, you have unique numbers to refer to in the paragraph

main/acle.md
```diff
@@ -1854,6 +1856,142 @@ or compiler-generated functions.
...
+
+### Name mangling
+
+The default version is not mangled top of the normal mangling.
```
fpetrogalli, line 1907: normal mangling? Is this the C++ mangling rules? If so, please specify.

main/acle.md
```diff
@@ -1854,6 +1856,142 @@ or compiler-generated functions.
...
+
+Version to be mangled. `<vendor-specific suffix>` should hold the version
```
fpetrogalli, line 1909: I think this sentence needs to be expanded a bit, and `<vendor-specific suffix>` should be defined (I cannot find its definition anywhere)

main/acle.md
```diff
@@ -1854,6 +1856,142 @@ or compiler-generated functions.
...
+information [[cxxabi]](#cxxabi) as:
+
+``` c
+    _(M<name>)*
+```
+name is from table below.
```
fpetrogalli, line 1915:
```suggestion
information [[cxxabi]](#cxxabi) as`_(M<name>)*`, where `<name>` is from the table below.
```

main/acle.md
```diff
@@ -1854,6 +1856,142 @@ or compiler-generated functions.
...
+
+If multiple features are requested then those shall be applied in priority order.
+
+### Mapping
+
+The following table lists the architecturesfeature mapping for AArch32
```
fpetrogalli, line 1921:
```suggestion
The following table lists the architecture feature mapping for AArch32.
```

main/acle.md
```diff
@@ -1854,6 +1856,142 @@ or compiler-generated functions.
...
+
+Version to be mangled. `<vendor-specific suffix>` should hold the version
+information [[cxxabi]](#cxxabi) as:
+
+``` c
+    _(M<name>)*
+```
+name is from table below.
+
+If multiple features are requested then those shall be applied in priority order.
```
fpetrogalli, line 1917:
```suggestion
If multiple features are requested then those shall be applied following rules specified in section [Selection](#selection).
```

main/acle.md
```diff
@@ -1854,6 +1856,142 @@ or compiler-generated functions.
...
+   | **Priority** | **Architecture name**            | **Name**        | **Dependent feature registers** |
+   | ------------ | -------------------------------- | --------------- | ------------------------------- |
+   | 0            | N/A                              | default         | ```N/A```                       |
+   | 90           | CRC32 instructions               | crc32           | ```ID_ISAR5.CRC32 == 0b0001```  |
+   | 100          | SHA1 instructions                | sha1            | ```ID_ISAR5.SHA1 == 0b0001```   |
+   | 110          | SHA2 instructions                | sha256          | ```ID_ISAR5.SHA2 == 0b0001```   |
+   | 120          | AES instructions                 | aes             | ```ID_ISAR5.AES >= 0b0001```    |
+   | 130          | VMULL (polynomial) instructions  | vmull           | ```ID_ISAR5.AES == 0b0002```    |
+   
+The following table lists the architecturesfeature mapping for AArch64
```
fpetrogalli, line 1932:
```suggestion
The following table lists the architecture feature mapping for AArch64.
```

main/design_documents/function-multi-versioning.md
```diff
@@ -0,0 +1,66 @@
...
+All toolchains and platforms should use the same source code formats to simplify
+the development and deployment of software and therefore increase the use of
+architectural features. Arm architectures are a bit more complex in view of
+feature availability therefore function versions would be better to be bound to
+features instead of cpu implementations. Also the kernels may limit the
+available features in the user space. Arm architectures are a bit more complex
+in view of feature availability therefore function versions would be better to
+be bound to features instead of cpu implementations. Also the kernels may limit
+the available features in the user space. Function Multi Versioning abbreviated
+to FMV.
```
fpetrogalli, line 15: Duplicated sentences?

main/design_documents/function-multi-versioning.md
```diff
@@ -0,0 +1,66 @@
+# Function Multi Versioning support for Arm architectures
+
+Other architectures supports [function multi versioning]
```
fpetrogalli, line 3:
```suggestion
Other architectures supports [Function Multi Versioning (FMV)]
```

main/design_documents/function-multi-versioning.md
```diff
@@ -0,0 +1,66 @@
...
+(https://gcc.gnu.org/wiki/FunctionMultiVersioning). This documents presents
+the rationale behind the design of the same feature for Arm.
+All toolchains and platforms should use the same source code formats to simplify
```
fpetrogalli, line 6:
```suggestion

All toolchains and platforms should use the same source code formats to simplify
```

main/design_documents/function-multi-versioning.md
```diff
@@ -0,0 +1,66 @@
...
+architectural features. Arm architectures are a bit more complex in view of
+feature availability therefore function versions would be better to be bound to
+features instead of cpu implementations. Also the kernels may limit the
+available features in the user space. Arm architectures are a bit more complex
+in view of feature availability therefore function versions would be better to
+be bound to features instead of cpu implementations. Also the kernels may limit
+the available features in the user space. Function Multi Versioning abbreviated
+to FMV.
+
+Why ACLE instead of ABI? FMV does not change how the function behaves from the
```
fpetrogalli, line 17:
```suggestion
Why is FMV defined in the ACLE instead of the ABI? Because FMV does not change how the function behaves from the
```

main/design_documents/function-multi-versioning.md
```diff
@@ -0,0 +1,66 @@
...
+ABI perspective. Because the caller does not know which version is actually
+called, the following holds for all the function derived from a function via
+FMV:
+
+1. the derived function have the same signature of the original function,
+2. the derived function obey to the same calling convention of the original
+function.
```
fpetrogalli, line 24: Aren't rule 1 and 2 something that should actually go into the specs?

main/design_documents/function-multi-versioning.md
```diff
@@ -0,0 +1,66 @@
...
+
+Today the `target` [attribute aarch64](https://gcc.gnu.org/onlinedocs/gcc/AArch64-Function-Attributes.html#AArch64-Function-Attributes>)
```
fpetrogalli, line 26:
```suggestion
Today the [`target` attribute for aarch64](https://gcc.gnu.org/onlinedocs/gcc/AArch64-Function-Attributes.html#AArch64-Function-Attributes>)
```

main/design_documents/function-multi-versioning.md
```diff
@@ -0,0 +1,66 @@
...
+is used for many purposes, some of which might overlap the functionality
+introduced by FMV. To avoid confusion, we propose the  introduction of the
+`target_version` attribute besides the `target_clones` attribute.
```
fpetrogalli, line 29:
```suggestion
introduced by FMV. To avoid confusion, we named the attributes used by FMV with
`target_version` and `target_clones`.
```

main/design_documents/function-multi-versioning.md
```diff
@@ -0,0 +1,66 @@
...
+
+The ACLE does not specify the underlying mechanism of the selection and the
```
fpetrogalli, line 31: Is this true? This seems to be defined in the `Selection` section of the specs?

main/design_documents/function-multi-versioning.md
```diff
@@ -0,0 +1,66 @@
...
+we propose the following feature test macro, to be
+defined to 1 if the support is available:
+
+``` c
+  __ARM_FEATURE_FUNCTION_MULTI_VERSIONING
+```
+
+Enabling a feature for given function does not enable the related ACLE
+functionalities like the intrinsics in the function's scope. That would require
+significant amount of changes in the C/C++ preprocessor.
```
fpetrogalli, line 52:
```suggestion
Enabling this feature and using one of the attributes on a given function does not enable the related ACLE feature(s) associated to the `"name"` expressed in the attribute.
 That would require
significant amount of changes in the C/C++ preprocessor.
```

main/design_documents/function-multi-versioning.md
```diff
@@ -0,0 +1,66 @@
...
+
+- The compiler is not aware of the new attributes.
+
+Usually unknown attributes are ignored, but the feature test macro is there to
+disable the unwanted code parts.
+
+- FMV aware compiler processes the attributes with newer arch features.
+
+In this case the compiler should ignore the new feature because it won't able to
+process it.
```
fpetrogalli, line 66:
```suggestion
These use cases to be considered:

* *The compiler is not aware of the new attributes introduced by FMV*: usually unknown attributes are ignored, but the feature test macro is there to
disable the unwanted code parts.

* *A compiler is aware of FMV but it encounters an attribute with and unknown architectural  features expressed in `<name>`*: in this case the compiler  will not be able to recognize the feature therefore it will be ignored.
```

fpetrogalli (changes requested):
hi @DanielKristofKiss - thank you for updating this to markdown.

I have few ore comments, that I think will make it easier for the reader to understand this feature.

I see you have  few comments from @rsandifo-arm / @nsz-arm / @enh-google.

May I ask you to make sure they formally approve the PR if they are happy with it, and mark those comments as resolved? I think you should make clear to them that we intend to release this feature in [2022Q1](https://github.com/ARM-software/acle/milestone/3) at BETA level of support.

Thank you!

Francesco

main/acle.md
DanielKristofKiss: ack

main/acle.md
DanielKristofKiss: ack.

main/acle.md
DanielKristofKiss: `<vendor-specific suffix>` comes from the cxxabi:
https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangling-structure

main/acle.md
DanielKristofKiss: Here I referred to the mangling, I'll add an example.

main/design_documents/function-multi-versioning.md
DanielKristofKiss: ack

main/design_documents/function-multi-versioning.md
DanielKristofKiss: ack

main/design_documents/function-multi-versioning.md
DanielKristofKiss: The mechanics of the selection is up to the implementation but expected results are defined in the `Selection` section.

main/acle.md
DanielKristofKiss: ack

main/acle.md
DanielKristofKiss: ack.

main/acle.md
DanielKristofKiss: ack

main/design_documents/function-multi-versioning.md
DanielKristofKiss: "These attributes have no effect on the calling convention, all versions must
use the same calling convention." is in the spec.

fpetrogalli (changes requested):
Just some minor stuff.

Francesco

main/design_documents/function-multi-versioning.md
DanielKristofKiss: yes

main/acle.md
```diff
@@ -1888,6 +1889,146 @@ or compiler-generated functions.
...
+
+1. These attributes can't be mixed with each other.
+2. The `"default"` version means the version of the function that would be generated
+without these attributes.
+3. `"name"` is the dependent features from the tables below.
+4. The dependent features could be joined by the `+` sign.
+5. None of these attributes will enable the corresponding ACLE feature(s) associated
+to the `"name"` expressed in the attribute.
+6. These attributes have no effect on the calling convention, all versions must
+use the same calling convention.
```
fpetrogalli, line 1913: This can be confusing, as there are attributes that can modify the calling convention [1] of a specific function without being applied to all version. I would state is as follows:

```suggestion
6. These attributes have no effect on the calling convention.
```

[1] See `aarch64_vector_pcs` at https://github.com/ARM-software/abi-aa/blob/main/vfabia64/vfabia64.rst#vector-procedure-call-standard

main/acle.md
```diff
@@ -1888,6 +1889,146 @@ or compiler-generated functions.
...
+
+`__ARM_FEATURE_FUNCTION_MULTI_VERSIONING` is defined to 1 if the versioning
+mechanism described in this section is available in the compiler and it is enabled.
+
+### Name mangling
+
+The `"default"` version is not mangled top of the languge specific name mangling.
+
+Function version to be mangled for C/C++. `<vendor-specific suffix>` should hold the version
+information [[cxxabi]](#cxxabi) as `_(M<name>)*` `name` is from tables below.
```
fpetrogalli, line 1944: This needs to be clarified. May I suggest the following change:

```suggestion

The mangling function is compatible with the mangling for version information of the [[cxxabi]](#cxxabi), and it  is defined as follows:

` ` ` 
<variant name> := <c/c++mangling> `.` <vendor specific suffix> 
<c/c++ mangling> := function name mangling for c/c++
<vendor specific suffix> := token obtained from the tables below
` ` `
```

main/acle.md
```diff
@@ -1888,6 +1889,146 @@ or compiler-generated functions.
...
+will produce these mangled names for C languae: `foo`, `foo._Mcrc32`, `foo._Msha1Maes`.
+
+
+### Mapping
+
+The following table lists the architectures feature mapping for AArch32
+
+   | **Priority** | **Architecture name**            | **Name**        | **Dependent feature registers** |
+   | ------------ | -------------------------------- | --------------- | ------------------------------- |
+   | 0            | N/A                              | default         | ```N/A```                       |
```
fpetrogalli, line 1962: I think that the triple ` is not needed here in this table... I think it would be best to get to single `

main/acle.md
```diff
@@ -1888,6 +1889,146 @@ or compiler-generated functions.
...
+   | 100          | SHA1 instructions                | sha1            | ```ID_ISAR5.SHA1 == 0b0001```   |
+   | 110          | SHA2 instructions                | sha256          | ```ID_ISAR5.SHA2 == 0b0001```   |
+   | 120          | AES instructions                 | aes             | ```ID_ISAR5.AES >= 0b0001```    |
+   | 130          | VMULL (polynomial) instructions  | vmull           | ```ID_ISAR5.AES == 0b0002```    |
+
+The following table lists the architectures feature mapping for AArch64
+
+   | **Priority**  | **Architecture name**    | **Name**      | **Dependent feature registers**           |
+   | ------------- | ------------------------ | ------------- | ----------------------------------------- |
+   | 0             | N/A                      | default       | N/A                                       |
```
fpetrogalli, line 1973: Sometimes you write N/A and sometimes you write `N/A`. I think we can stick with the former (no ticks).

main/acle.md
```diff
@@ -1888,6 +1889,146 @@ or compiler-generated functions.
...
+   | 420           | `FEAT_SSBS2`             | ssbs2         | ```ID_AA64PFR1_EL1.SSBS == 0b0010```      |
+   | 430           | `FEAT_BTI`               | bti           | ```ID_AA64PFR1_EL1.bt == 0b0010```        |
+
+### Selection
+
+The following rules shall be followed by all implementations. Implementation of
+the selection algorithm is platform dependent, where with platform means
+CPU/Vendor/OS as in the target triplet. The selection is permanent for the
+lifetime of the process. Only those versions could be considered where all
+dependent features are available.
```
fpetrogalli, line 2024: I think we need itemization here too...

```suggestion
The following rules shall be followed by all implementations.

1. Implementation of
the selection algorithm is platform dependent, where with platform means
CPU/Vendor/OS as in the target triplet. 
2. The selection is permanent for the
lifetime of the process.
3. Only those versions could be considered where all
dependent features are available.
```

main/acle.md
```diff
@@ -1888,6 +1889,146 @@ or compiler-generated functions.
...
+
+Rules of version selection are in order:
+
+1. Select the most specific version else
+2. select the version with the highest priority else
+3. `"default"` shall be selected if no other versions are suitable.
```
fpetrogalli, line 2030: ... and consequently, given we have already used 1,2,3 in the previous paragraph, we might want to numbers these as 4 5 and 6.

main/acle.md
```diff
@@ -1888,6 +1889,146 @@ or compiler-generated functions.
...
+version shall be used.
+
+The attribute `__attribute__((target_version("name")))` expresses the following:
+
+1. when applied to a function it becomes one of the versions. Function with the same
+name may exist with multiple versions in the same translation unit.
+2. One `"default"` version of the function is required to be provided.
+   a. Implicitly, without this attribute,
+   b. or explicitly providing the `"default"` in the attribute.
+4. All instances of the versions shall share the same function signature.
```
fpetrogalli, line 1925: Inside this section I'd number these as 7. 8. 9. to be able to uniquely reference them...

main/acle.md
```diff
@@ -1888,6 +1889,146 @@ or compiler-generated functions.
...
+
+The attribute `__attribute__((target_clones("name",...)))` expresses the following:
+
+1. when applied to a function the compiler emits multiple versions based on the
+arguments.
+   a. One of them is implicitly the `"default"`.
+   b. If the `"default"` matches with an other explicitly provided version the
+compiler can emit only one function instead of the two.
+2. If a name is not recognized the compiler should ignore it.
```
fpetrogalli, line 1934: ...and consequently number these as 10. 11.

fpetrogalli (changes requested):
I read the whole thing once more and added few more comments that I think it would improve the way the content is presented.

Francesco

main/acle.md
DanielKristofKiss: I'd like to keep the second part as a separate item. Since the callee calls all versions in the same way changing only one  calling convention one of the function won't work.

main/acle.md
DanielKristofKiss: `_` separator and M prefix for tokeniser is missing but I apply this format.

main/acle.md
```diff
@@ -1888,6 +1889,163 @@ or compiler-generated functions.
...
+
+If multiple features are requested then those shall be appended in priority
+order and prefixed with `M`.
+
+For example:
+``` c
+__attribute__((target_clones("crc32", "aes+sha1")))
+int foo(){..}
+```
+will produce these mangled names for C languae: `foo`, `foo._Mcrc32`,
```
fpetrogalli, line 1964:
```suggestion
will produce these mangled names for C language: `foo`, `foo._Mcrc32`,
```

fpetrogalli (changes requested):

fpetrogalli (approved):
Thank you @DanielKristofKiss

fpetrogalli:
@allcontributors please add @DanielKristofKiss  for code.

echristo:
Can we merge the commits here to make it a little easier for review?

PR #72, opened 2021-10-25
Add NEON-SVE Bridge

MDevereau:
Adds section 1.11.1 NEON-SVE Bridge which details the addition of
svextract_ and svinsert_

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -194,6 +194,8 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * Added reference to the *Cortex-M Security Extension (CMSE)*
   specifications in
   [sec-CMSE-intrinsics](#cortex-m-security-extension-cmse).
+* Added specification for [sec-NEON-SVE-Bridge](#neon-sve-bridge) 
+  and [sec-NEON-SVE-Bridge-macros](#neon-sve-bridge-macros).
 
 ### References
 
@@ -1647,6 +1649,11 @@ be found in [[BA]](#BA).
 | `__ARM_SIZEOF_MINIMAL_ENUM` [[ssec-Imptype]](#implementation-defined-type-properties)                                    | Size of minimal enumeration type: 1 or 4                                                           | 1           |
 | `__ARM_SIZEOF_WCHAR_T` [[ssec-Imptype]](#implementation-defined-type-properties)                                         | Size of `wchar_t`: 2 or 4                                                                          | 2           |
 | `__ARM_WMMX` [[ssec-WMMX]](#wireless-mmx)                                                         | Wireless MMX extension (32-bit-only)                                                               | 1           |
+
+## NEON-SVE Bridge macros
+
+`__ARM_NEON_SVE_BRIDGE` is defined to 1 if [NEON-SVE-Bridge](#neon-sve-bridge)
+intrinsics are available.
 
 # Attributes and pragmas
 
@@ -5064,3 +5071,91 @@ transaction.
 | uint64_t __ttest (void)                       | -                | Xt -> result   | ttest <Xt>        |
 
 These intrinsics are available when `arm_acle.h` is included.
+
+# Architectural Extension Bridges
+
+## NEON-SVE Bridge
+
+The NEON_SVE Bridge adds intrinsics that allow conversions between NEON and
+SVE vectors.
+
+These intrinsics are defined in the header file `arm_neon_sve_bridge.h`.
+
+`__ARM_NEON_SVE_BRIDGE` is defined as 1 when `arm_neon_sve_bridge.h` is
+available.
+
+Including this header_file will include `arm_neon.h` and `arm_sve.h`.
+
+### `svset_neonq`
+
+These intrinsics set the first 128 bits of SVE vector `vec` to `subvec`.
+
+That is, bit *i* of the result is equal to:
+* bit *i* of `subvec` if *i* < 128
+* bit *i* of `vec` otherwise
+
+On big-endian targets, this leaves lanes in a different
+order from the “native” SVE order.  For example, if `subvec` is
+`int32x4_t`, then on big-endian targets, the first memory element
+is in lane 3 of `subvec` and is therefore in lane 3 of the returned
+SVE vector.  Using `svld1` to load elements would instead put the
+first memory element in lane 0 of the returned SVE vector.
+
+When `svundef` is passed as the `vec` parameter, compilers are able
+to reuse the SVE register overlapping the NEON input without generating
+additional instructions.
+
+| **Instances**                                                            |
+|--------------------------------------------------------------------------|
+| `svint8_t svset_neonq[_s8](svint8_t vec, int8x16_t subvec)`              |
+| `svint16_t svset_neonq[_s16](svint16_t vec, int16x8_t subvec)`           |
+| `svint32_t svset_neonq[_s32](svint32_t vec, int32x4_t subvec)`           |
+| `svint64_t svset_neonq[_s64](svint64_t vec, int64x2_t subvec)`           |
+| `svuint8_t svset_neonq[_u8](svuint8_t vec, uint8x16_t subvec)`           |
+| `svuint16_t svset_neonq[_u16](svuint16_t vec, uint16x8_t subvec)`        |
+| `svuint32_t svset_neonq[_u32](svuint32_t vec, uint32x4_t subvec)`        |
+| `svuint64_t svset_neonq[_u64](svuint64_t vec, uint64x2_t subvec)`        |
+| `svfloat16_t svset_neonq[_f16](svfloat16_t vec, float16x8_t subvec)`     |
+| `svfloat32_t svset_neonq[_f32](svfloat32_t vec, float32x4_t subvec)`     |
+| `svfloat64_t svset_neonq[_f64](svfloat64_t vec, float64x2_t subvec)`     |
+| `svbfloat16_t svset_neonq[_bf16](svbfloat16_t vec, bfloat16x8_t subvec)` |
+
+### `svget_neonq`
+
+These intrinsics get the first 128 bit subvector of SVE vector `vec` as a
+NEON vector.
+
+| **Instances**                                       |
+|-----------------------------------------------------|
+| `int8x16_t svget_neonq[_s8](svint8_t vec)`          |
+| `int16x8_t svget_neonq[_s16](svint16_t vec)`        |
+| `int32x4_t svget_neonq[_s32](svint32_t vec)`        |
+| `int64x2_t svget_neonq[_s64](svint64_t vec)`        |
+| `uint8x16_t svget_neonq[_u8](svuint8_t vec)`        |
+| `uint16x8_t svget_neonq[_u16](svuint16_t vec)`      |
+| `uint32x4_t svget_neonq[_u32](svuint32_t vec)`      |
+| `uint64x2_t svget_neonq[_u64](svuint64_t vec)`      |
+| `float16x8_t svget_neonq[_f16](svfloat16_t vec)`    |
+| `float32x4_t svget_neonq[_f32](svfloat32_t vec)`    |
+| `float64x2_t svget_neonq[_f64](svfloat64_t vec)`    |
+| `bfloat16x8_t svget_neonq[_bf16](svbfloat16_t vec)` |
+
+### `svdup_neonq`
+
+These intrinsics return an SVE vector with all SVE subvectors containing the
+duplicated NEON vector `vec`.
+
+| **Instances**                                       |
+|-----------------------------------------------------|
+| `svint8_t svdup_neonq[_s8](int8x16_t vec)`          |
+| `svint16_t svdup_neonq[_s16](int16x8_t vec)`        |
+| `svint32_t svdup_neonq[_s32](int32x4_t vec)`        |
+| `svint64_t svdup_neonq[_s64](int64x2_t vec)`        |
+| `svuint8_t svdup_neonq[_u8](uint8x16_t vec)`        |
+| `svuint16_t svdup_neonq[_u16](uint16x8_t vec)`      |
+| `svuint32_t svdup_neonq[_u32](uint32x4_t vec)`      |
+| `svuint64_t svdup_neonq[_u64](uint64x2_t vec)`      |
+| `svfloat16_t svdup_neonq[_f16](float16x8_t vec)`    |
+| `svfloat32_t svdup_neonq[_f32](float32x4_t vec)`    |
+| `svfloat64_t svdup_neonq[_f64](float64x2_t vec)`    |
+| `svbfloat16_t svdup_neonq[_bf16](bfloat16x8_t vec)` |
--- a/main/design_documents/ARM_NEON_SVE_BRIDGE.md
+++ b/main/design_documents/ARM_NEON_SVE_BRIDGE.md
@@ -1,0 +1,67 @@
+# Design Document for ARM_NEON_SVE_BRIDGE
+
+This document describes the design of the intrinsics defined in
+[NEON_SVE Bridge](../acle.md#neon-sve-bridge). These C intrinsics enable the
+conversion between NEON and SVE vectors with no performance overhead, and
+duplicating NEON vectors across SVE vectors.
+
+## Location
+
+The intrinsics are defined in `arm_neon_sve_bridge.h`. If
+`arm_neon_sve_bridge.h` is available on the compiler,
+`__ARM_NEON_SVE_BRIDGE` is defined as 1. Users include the NEON/SVE Bridge
+as follows:
+
+```
+#ifdef __ARM_NEON_SVE_BRIDGE
+#include <arm_neon_sve_bridge.h>
+#endif //__ARM_NEON_SVE_BRIDGE
+```
+
+`arm_neon_sve_bridge.h` includes `arm_neon.h` and `arm_sve.h`.
+
+## Suggestions & Considerations
+
+### Why not enable the extraction of any 128-bit subvector in an SVE vector?
+
+Extracting other 128-bit subvectors goes against the no-op design.
+This functionality is already available with the ACLE intrinsic 
+`svdupq_lane`. For example:
+
+```
+svuint64_t sve_vector_a = ...;
+uint64x2_t neon_vec = svget_neonq_u64(svdupq_lane_u64(sve_vector_a, 1));
+```
+
+### Why not enable the insertion of a 128-bit NEON vector into a specific SVE subvector?
+
+`arm_sve.h` does not currently provide any intrinsics for inserting individual
+elements into a specific lane of an SVE vector. The approach should be
+consistent between `arm_sve.h` and `arm_neon_sve_bridge.h`
+
+### Why not include 64-bit Vectors?
+
+Support for 64-bit vectors is not added because of low interest. The
+naming scheme is designed to cope with the possibility of supporting these in
+the future. The C standard requires a unique function name for each set of
+input and return types from these functions. C Generics enable inferring the
+intended function from its parameter types, but not from the return type.
+Therefore we adopt the NEON convention of using the letter `q` in the name to
+indicate a 128-bit vector in the return type. e.g. `svget_neonq[_u16]`. If
+64-bit vectors are added later they will omit the `q`.
+
+#### NEON Vector duplication across all SVE vectors
+
+`svdup_neonq` is added as a convenience function to avoid casting from
+NEON to SVE as a separate step when duplicating NEON vectors into SVE vectors.
+`svdup_neonq` also prevents initializing non-NEON bits in an SVE vector
+and then immediately overriding them with an `svdupq_lane` intrinsic.
+
+### Intrinsic Naming
+
+The intrinsics are named `svset_neonq`, `svget_neonq` and
+`svdup_neonq`. These names are taken from `svget/svset[2|3|4]` and 
+`svdupq` from the [SVE ACLE specification](https://developer.arm.com/documentation/100987/).
+The `q` is from the 64/128 bit naming convention of the 
+[NEON](https://github.com/ARM-software/acle/tree/main/neon_intrinsics/)
+intrinsics.


main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,44 @@
+=======================================
+Design Document for ARM_NEON_SVE_BRIDGE
+=======================================
+
+This document describes the design of the intrinsics defined in the ("NEON_SVE bridge")
+<../acle.rst#_ssec-NEONSVEBridge>. These intrinics add a
```
MDevereau, line 6: After a quick playaround i've not managed to get this to link to `.. _ssec-NEONSVEBridge:` at line 411 in `main/acle.rst`

fpetrogalli: Github does not use the internal hyperlink names, but has it's own convention: https://github.com/MDevereau/acle/blob/next-release/main/acle.rst#neon-sve-bridge 

```suggestion
<../acle.rst#-neon-sve-brige>. These intrinsics add a
```

(mind a typo too)

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,44 @@
...
+
+Suggestions & Considerations
+############################
+
+Why not enable the extraction of any 128-bit subvector in an SVE vector?
+------------------------------------------------------------------------
+
+Extracting other 128-bit subvectors would go against the no-op design.
+There are currently other methods of getting data this way from SVE
+subvectors using SVE intrinsics.
```
georges-arm, line 30: Could you add an example of how to extract another 128-bit segment from an SVE vector? I am assuming the ideal codegen here would be `MOV <Zd>.<T>, <Zn>.<T>[<imm>]` but can't think of a sequence that would produce this off the top of my head?

main/acle.rst
```diff
@@ -404,6 +408,131 @@ Scalable Vector Extensions (SVE)
...
+and must be included after ``arm_neon.h`` and ``arm_sve.h``. These
+intrinsics are available in the ``aarch64`` state.
+
+Including this header file will define ``__ARM_SVE_NEON_BRIDGE`` as 1.
+
+``svset_neonq``
+~~~~~~~~~~~~~~~
+
+These intrinsics set NEON vector ``subvec`` into the first 128 bit subvector of
+the SVE vector ``vec``.
```
georges-arm, line 431: Is it worth including a version of these to set the low 128-bits of an SVE vector and ignore the top bits entirely. I imagine we can do that with this existing proposal with e.g. `svset_neonq_s8(svundef_s8(), a)` but I don't know whether there is a preference for relying on compilers optimising that away into a no-op versus providing a dedicated set of intrinsics to handle this?

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,44 @@
...
+
+Why not include 64-bit Vectors?
+-------------------------------
+
+64-bit vectors will not initially be added. The naming scheme of the intrinsics
+must respect possible future 64-bit additions. Following the NEON naming 
+convention, e.g. ``svget_neonq[_u16]`` and ``svset_neonq[_u16]`` resolves this.
```
georges-arm, line 37: I believe the `q` part of the name could also be inferred from the argument even in the presence of a future 64-bit vector extension? Don't have a strong preference for making it implicit or not, just pointing out that it is possible.

main/acle.rst
```diff
@@ -404,6 +408,131 @@ Scalable Vector Extensions (SVE)
...
+NEON-SVE Bridge
+---------------
+
+The NEON-SVE Bridge adds intrinsics that enables setting NEON vectors
+into SVE vectors, and getting NEON vectors from SVE vectors. As the first 128
+bit subvector of an SVE vector is physically a NEON register,
+these intrinsics result in a no-op.
+
+These intrinsics are defined in the header file ``arm_extensions_bridge.h``
+and must be included after ``arm_neon.h`` and ``arm_sve.h``. These
```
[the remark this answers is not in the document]
peterwaller-arm: Good point. How about an approach along the lines of forward declaring the types used by the bridge so that the header stands alone? It may be a small thing but it doesn't seem ideal to `#include` the whole header multiple times into the same compilation unit if it is not necessary.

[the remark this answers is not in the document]
MDevereau: I was thinking more along the lines of
`#define __ARM_NEON_SVE_BRIDGE (__ARM_FEATURE_NEON && __ARM_FEATURE_SVE)``.
``__ARM_FEATURE_NEON` rather than including the header files, but yes the forced order isnt ideal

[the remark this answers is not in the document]
MDevereau: or something like
```
#if !defined(__ARM_FEATURE_NEON) || !defined(__ARM_FEATURE_SVE)
  error
#else
  #define __ARM_NEON_SVE_BRIDGE 1
```

main/acle.rst
MDevereau: `svset_neonq_s8(svundef_s8(), a)` should be able to handle this perfectly fine we think.

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
MDevereau: I'm updating the spec to expand on this. Basically C generics can't infer 64/128 bits from the return type alone (only the parameters) so the function name needs to reflect this

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
MDevereau: `svdupq_lane_s8(x, 1);` compiles to `mov     z0.q, z0.q[1]` on the head of llvm

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,51 @@
+=======================================
+Design Document for ARM_NEON_SVE_BRIDGE
+=======================================
+
+This document proposes a new set of intrinsics. These intrinics add a
```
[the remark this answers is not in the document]
MDevereau: done

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,55 @@
...
+The intrinsics will be in a new header file called ``arm_extensions_bridge.h``.
+
+This header file will define ``__ARM_NEON_SVE_BRIDGE`` as 1.
+
+The intrinsics will be conditionally defined with
+
+.. code:: c
+
+    #if !defined(__ARM_FEATURE_NEON) || !defined(__ARM_FEATURE_SVE)
+        #error "arm_extensions_bridge.h requires -march=+simd+sve"
```
peterwaller-arm, line 22: Missing the #endif?

MDevereau: whoops. sorted

main/acle.rst
```diff
@@ -404,6 +408,131 @@ Scalable Vector Extensions (SVE)
...
 document [SVE-ACLE]_ available on the Arm Developer Website.
 
+.. _ssec-NEONSVEBridge:
+
+NEON-SVE Bridge
+---------------
+
+The NEON-SVE Bridge adds intrinsics that enables setting NEON vectors
+into SVE vectors, and getting NEON vectors from SVE vectors. As the first 128
+bit subvector of an SVE vector is physically a NEON register,
```
peterwaller-arm, line 418:
```suggestion
bit subvector of an SVE vector physically overlaps with a NEON register,
```
('is' is a strong word, and the two register classes are distinct)

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,56 @@
...
+supported in the aarch32 state.
+
+Suggestions & Considerations
+############################
+
+Why not enable the extraction of any 128-bit subvector in an SVE vector?
+------------------------------------------------------------------------
+
+Extracting other 128-bit subvectors would go against the no-op design.
+This functionality is already possible with the ACLE intrinsic `svdupq_lane`.
```
peterwaller-arm, line 37:
```suggestion
This functionality is already available with the ACLE intrinsic `svdupq_lane`.
```

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,56 @@
...
+
+Why not include 64-bit Vectors?
```
peterwaller-arm, line 39: The question is not directly answered (the "why") by the below paragraph.

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,56 @@
+=======================================
+Design Document for ARM_NEON_SVE_BRIDGE
+=======================================
+
+This document describes the design of the intrinsics defined in the ("NEON_SVE bridge")
+<../acle.rst#neon-sve-bridge>. These intrinsics add a
+convienient way to use NEON and SVE vectors interchangeably in C code with
```
peterwaller-arm, line 7: On a strict reading of this, you cannot interchange an SVE and a NEON register in an arbitrary context, so I think you intend to write something different here.
```suggestion
convienient way to convert between NEON and SVE vectors in C code with
```

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
MDevereau: The direct answer is just that there isn't enough interest currently right?

main/acle.rst
```diff
@@ -404,6 +408,131 @@ Scalable Vector Extensions (SVE)
...
+|      svfloat16_t svset_neonq[_f16](svfloat16_t vec, float16x8_t subvec)        |
++--------------------------------------------------------------------------------+
+| .. code:: c                                                                    |
+|                                                                                |
+|      svfloat32_t svset_neonq[_f32](svfloat32_t vec, float32x4_t subvec)        |
++--------------------------------------------------------------------------------+
+| .. code:: c                                                                    |
+|                                                                                |
+|      svfloat64_t svset_neonq[_f64](svfloat64_t vec, float64x2_t subvec)        |
++--------------------------------------------------------------------------------+
```
georges-arm, line 479: Can you also add instances of both get and set for the `poly8x16_t` , `poly16x8_t` , `poly64x2_t` and `poly128_t` types?

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,56 @@
...
+The intrinsics will be in a new header file called ``arm_extensions_bridge.h``.
+
+This header file will define ``__ARM_NEON_SVE_BRIDGE`` as 1.
+
+The intrinsics will be conditionally defined with
+
+.. code:: c
+
+    #if !defined(__ARM_FEATURE_NEON) || !defined(__ARM_FEATURE_SVE)
+        #error "arm_extensions_bridge.h requires -march=+simd+sve"
```
georges-arm, line 22: Not sure this is the right error to give here. Trying to use this option with GCC gives `cc1: error: missing arch name in '-march=+simd+sve'`. Clang gives `clang-13: error: the clang compiler does not support '-march=+simd+sve'`.

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,56 @@
...
+    #endif
+    #define __ARM_NEON_SVE_BRIDGE 1
+    //define intrinsics
+
+`__ARM_FEATURE_NEON` is used as it is aarch64 specific, and SVE is not
```
georges-arm, line 27: Capitalisation: "AArch64" / "AArch32"?

main/acle.rst
MDevereau: Any suggestions for the `poly128_t` signature? There's no `svint128_t`

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
MDevereau: `requires -march=armv8-a+simd+sve`?

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
MDevereau: done

main/acle.rst
georges-arm: I didn't notice that the polynomial types don't exist anymore! I would guess probably the corresponding unsigned types should be used (to match the `pmull{b,t}` intrinsics), with `poly128_t` corresponding to `svuint64x2_t` based on the signature of e.g. `svpmullb_pair`? Would prefer a more authoritative opinion though, @fpetrogalli?

main/acle.rst
MDevereau: I've removed this sentence entirely after an internal discussion

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
MDevereau: done

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
MDevereau: done

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
MDevereau: If Neon support is mandatory in AArch64 I wonder if theres any point in me asserting against `__ARM_FEATURE_NEON` anywhere at all

MDevereau:
```
int main(){
#ifdef __ARM_FEATURE_NEON
        printf("neon enabled\n");
#endif
#ifdef __ARM_FEATURE_SVE
        printf("sve enabled\n");
#endif
        return 0;
}
```
Compiling the above with the trunk of clang with `-march=armv8-a+simd+sve` only outputs `sve enabled`... interesting

MDevereau: I'm guessing i should just be using `__ARM_NEON` then?

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
peterwaller-arm: Yes.

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
peterwaller-arm: > If Neon support is mandatory in AArch64 I wonder if theres any point in me asserting against `__ARM_FEATURE_NEON` anywhere at all

Good thought - it's difficult to imagine a scenario where the NEON feature is disabled but SVE is enabled, so I believe testing for SVE should be sufficient.

MDevereau: There's still the problem that SVE doesn't bring in neon types and we still need to be sure we have the neon types defined when we include this file

peterwaller-arm: I thought the intent was to resolve that issue by forward declaring the neon types?

main/acle.rst
```diff
@@ -404,6 +408,129 @@ Scalable Vector Extensions (SVE)
...
 
+.. _ssec-NEONSVEBridge:
+
+NEON-SVE Bridge
+---------------
+
+The NEON_SVE Bridge adds intrinsics that allow conversions between NEON and
+SVE vectors.
+
+These intrinsics are defined in the header file ``arm_extensions_bridge.h``.
```
rsandifo-arm, line 419: Perhaps this should be called something like arm_sve_neon.h or arm_neon_sve.h instead.  I think it will/should include both arm_neon.h and arm_sve.h.

The problem with a general header file like arm_extensions_bridge.h is that, over time, the header might accrue other unrelated features that the user wasn't expecting, possibly “infecting” the user's namespace with unforeseen styles of identifier.  For xample, suppose “arm_extensions_bridge.h” already existed for an earlier feature and that we then added the Neon/SVE bridge to it.  Users of the original version who upgrade their toolchain might find that arm_neon.h and arm_sve.h are now being included unexpectedly.

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
MDevereau: https://github.com/ARM-software/acle/pull/72#discussion_r740912334 we mentioned it here but didnt explicitly agree on it. From that conversation I was under the impression we were checking the `__ARM_FEATURE` macros to ensure the types had already been defined

main/acle.rst
rsandifo-arm: Personally, I think we should only provide functions for element types that exist in both Neon and SVE.  The user can reinterpret the Neon type where necessary.

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
peterwaller-arm: Ah, last I saw (and we discussed), you said 'forward declaring the types is probably the best bet' which sounded good to me, since it makes the order of includes not matter, which is important in the face of tools which might reorder the headers and helps with the principle of least surprise for users. I'll replied there too.

main/acle.rst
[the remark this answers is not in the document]
peterwaller-arm: Forward-declaring the types sounds like the most palatable solution to me at the moment to make the headers work regardless of order (principle of least surprise for the user), and to avoid double-including them or anything like that.

rsandifo-arm (commented):
Thanks for doing this.  A couple of general comments:

* I agree with George that it might be worth having convenience functions for “extend with undef”.  “Extend with zero” would be useful too, since setting a Neon register implicitly zero extends.  The justification in both cases is that, if an explicit SVE input is required, even the “overloaded” function style would require an explicit type suffix on any svundef or svdup call.  Having convenience functions would allow the call, and thus the type suffix, to be omitted.  This is a bit like the way that we provide _n versions of most arithmetic functions, even though they could be written using svdup instead.
* It might be worth having “svdup_neonq” functions to duplicate a neon vector to fill an SVE one.  This is just another way of writing svdupq_lane, without having to cast from Neon to SVE as a separate step.  I see you've considered and rejected the possibility of _extracting_ a Neon vector from an arbitrary point in the vector, and I think that's fair.  The reasons I think the dup functions are nevertheless worthwhile are that:
  * without it, the user would have to explicitly choose how to initialise the upper bits of the SVE register (in the svset_neonq call) only to overwrite those bits immediately.
  * providing the function is consistent with svdup for scalar floating-point values, which are likewise implemented using the lane form of DUP.

main/acle.rst
peterwaller-arm: > Perhaps this should be called something like arm_sve_neon.h or arm_neon_sve.h instead. I think it will/should include both arm_neon.h and arm_sve.h.
If it does include those headers internally, will this not enlarge the total bytes processed by the compiler a fair amount? In total for clang this is 4MiB of source. If this header includes it, and the user includes it, then the compiler is going to have to pull in 8MiB of source. Even if this is not fully parsed it does represent a fair amount of extra work for every compilation unit which might include this header. I don't know for sure if this is worth being concerned about, but it seems we can make that problem go away if the new header forward declares the types it needs internally. Do you have any opinions on this problem and proposed solution?

rsandifo-arm: I think it's fair to assume that people will only use the Neon/SVE bridge if they are also using Neon and SVE individually.  So from a user-interface perspective, I think we should document in the ACLE that including the new header file also has the effect of including arm_neon.h and arm_sve.h (just like including arm_sve.h is defined to have the effect of including arm_bf16.h when bfloat16 is supported).  Users can then write:

```
#include <new_header.h>
```

rather than having to write:

```
#include <arm_neon.h>
#include <arm_sve.h>
#include <new_header.h>
```

Even if they do (redundantly) include all three, I think the preprocessor has an optimisation that avoids reading a file a second time if all non-writespace/comment characters were protected by a macro guard.  (At least, GCC does this.  I assume Clang does too. :-))

peterwaller-arm: 👍 thanks for the input, SGTM.

main/acle.rst
MDevereau: I've decided to not to include polynomial types for now

MDevereau:
@rsandifo-arm Thank you for the review. I've added `svdup_neonq` for duplicating a NEON vector across all subvectors of an SVE vector. I've also changed the header files name from `arm_extensions_bridge.h` to `arm_neon_sve_bridge.h`

main/acle.rst
MDevereau: I've changed the header file's name from `arm_extensions_bridge.h` to `arm_neon_sve_bridge.h`. I've updated the spec to say it now includes `arm_neon.h` and `arm_sve.h`

main/acle.rst
```diff
@@ -404,6 +408,182 @@ Scalable Vector Extensions (SVE)
...
+|                                                              |
+|      int32x4_t svget_neonq[_s32](svint32_t vec)              |
++--------------------------------------------------------------+
+| .. code:: c                                                  |
+|                                                              |
+|      int64x2_t svget_neonq[_s64](svint64_t vec)              |
++--------------------------------------------------------------+
+| .. code:: c                                                  |
+|                                                              |
+|      uint8x16_t svget_neonq[_s8](svuint8_t vec)              |
```
pthariensflame, line 506: Typo here and below:  `s` should be `u`.

pthariensflame, line 506:
```suggestion
|      uint8x16_t svget_neonq[_u8](svuint8_t vec)              |
```

main/acle.rst
```diff
@@ -404,6 +408,182 @@ Scalable Vector Extensions (SVE)
...
++--------------------------------------------------------------+
+| .. code:: c                                                  |
+|                                                              |
+|      uint16x8_t svget_neonq[_s16](svuint16_t vec)            |
```
pthariensflame, line 510:
```suggestion
|      uint16x8_t svget_neonq[_u16](svuint16_t vec)            |
```

main/acle.rst
```diff
@@ -404,6 +408,182 @@ Scalable Vector Extensions (SVE)
...
++--------------------------------------------------------------+
+| .. code:: c                                                  |
+|                                                              |
+|      uint32x4_t svget_neonq[_s32](svuint32_t vec)            |
```
pthariensflame, line 514:
```suggestion
|      uint32x4_t svget_neonq[_u32](svuint32_t vec)            |
```

main/acle.rst
```diff
@@ -404,6 +408,182 @@ Scalable Vector Extensions (SVE)
...
++--------------------------------------------------------------+
+| .. code:: c                                                  |
+|                                                              |
+|      uint64x2_t svget_neonq[_s64](svuint64_t vec)            |
```
pthariensflame, line 518:
```suggestion
|      uint64x2_t svget_neonq[_u64](svuint64_t vec)            |
```

pthariensflame (changes requested):

main/acle.rst
MDevereau: Thanks :)

pthariensflame (approved):

main/acle.rst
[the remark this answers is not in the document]
fpetrogalli: Resolving this conversation as I understand we have decided to force the inclusion of `arm_sve.h` and `arm_neon.h` directly in the header file of the bridge.

main/acle.rst
```diff
@@ -273,6 +273,10 @@ Changes between ACLE Q2 2021 and ACLE Q3 2021
 * Introduce ``__ARM_FEATURE_PAUTH`` and ``__ARM_FEATURE_BTI`` in sections
   ssec-PAC_ and ssec-BTI_ respectively.
 
+Changes for next-release
+~~~~~~~~~~~~~~~~~~~~~~~~
+* Add specification for NEON-SVE Bridge.
```
fpetrogalli, line 278: Please link the text of NEON-SVE Bridge to the `_ssec-NEONSVEBridge:` anchor.

main/acle.rst
fpetrogalli: May I ask you to add this explanation in the design document?

```
Why polynomial type are not supported
---------------------

We only provide functions for element types that exist in both Neon and SVE.
The user can reinterpret the Neon type where necessary.

```

main/acle.rst
```diff
@@ -404,6 +408,182 @@ Scalable Vector Extensions (SVE)
...
+| .. code:: c                                                  |
+|                                                              |
+|      float32x4_t svget_neonq[_f32](svfloat32_t vec)          |
++--------------------------------------------------------------+
+| .. code:: c                                                  |
+|                                                              |
+|      float64x2_t svget_neonq[_f64](svfloat64_t vec)          |
++--------------------------------------------------------------+
+
+``svdup_neonq``
```
fpetrogalli, line 533: I think we need a section in the design document explaining why we are adding these.

@rsandifo-arm wrote:

> It might be worth having “svdup_neonq” functions to duplicate a neon vector to fill an SVE one. This is just another way of writing svdupq_lane, without having to cast from Neon to SVE as a separate step. I see you've considered and rejected the possibility of extracting a Neon vector from an arbitrary point in the vector, and I think that's fair. The reasons I think the dup functions are nevertheless worthwhile are that:
>    * without it, the user would have to explicitly choose how to initialise the upper bits of the SVE register (in the svset_neonq call) only to overwrite those bits immediately.
>    * providing the function is consistent with svdup for scalar floating-point values, which are likewise implemented using the lane form of DUP.

Please summarize this in a new subsection in the design document.

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,64 @@
...
+
+This document describes the design of the intrinsics defined in
+("NEON_SVE bridge")<../acle.rst#neon-sve-bridge>. These C intrinsics enable the
+conversion between NEON and SVE vectors with no performance overhead, and
+duplicating NEON vectors across SVE vectors.
+
+Location
+########
+
+The intrinsics will be defined in a new header file called
```
fpetrogalli, line 13:
```suggestion
The intrinsics is be defined in a header file called 
```

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,64 @@
...
+duplicating NEON vectors across SVE vectors.
+
+Location
+########
+
+The intrinsics will be defined in a new header file called 
+``arm_neon_sve_bridge.h``.
+
+This header file will include ``arm_neon.h`` and ``arm_sve.h``.
+It will also define ``__ARM_NEON_SVE_BRIDGE`` as 1.
```
fpetrogalli, line 17:
```suggestion
This header file includes ``arm_neon.h`` and ``arm_sve.h``.
It also defines ``__ARM_NEON_SVE_BRIDGE`` as 1.
```

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,64 @@
...
+
+The intrinsics will be defined with:
```
fpetrogalli, line 19:
```suggestion
The intrinsics are defined with:
```

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,64 @@
...
+    #ifdef __ARM_NEON_SVE_BRIDGE 1
+    //define intrinsics
+
+Suggestions & Considerations
+############################
+
+Why not enable the extraction of any 128-bit subvector in an SVE vector?
+------------------------------------------------------------------------
+
+Extracting other 128-bit subvectors would go against the no-op design.
```
fpetrogalli, line 32:
```suggestion
Extracting other 128-bit subvectors goes against the no-op design.
```

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,64 @@
...
+
+``svuint64_t sve_vector_a = ...;``
+
+``uint64x2_t neon_vec = svget_neonq_u64(svdupq_lane_u64(sve_vector_a, 1));``
+
+Why not enable the insertion of a 128-bit NEON vector into a specific SVE subvector?
+------------------------------------------------------------------------------------
+
+This behaviour is not currently available with SVE intrinsics. Such an intrinsic
+would be better placed in the SVE specification. NEON/SVE Bridge only concerns
```
fpetrogalli, line 44:
```suggestion
is better placed in the SVE specification. NEON/SVE Bridge only concerns 
```

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,64 @@
...
+Therefore we adopt the NEON convention of using the letter `q` in the name to
+indicate a 128-bit vector in the return type. e.g. ``svget_neonq[_u16]``. If
+64-bit vectors are added later they will omit the `q`.
+
+Intrinsic Naming
+----------------
+
+The intrinsics will be named ``svset_neonq``, ``svget_neonq`` and
+``svdup_neonq``. These names are taken from ``svget/svset[2|3|4]``. and 
+``svdupq``. The ``q`` is from the NEON 64/128 bit naming convention.
```
fpetrogalli, line 64:
```suggestion
The intrinsics are named ``svset_neonq``, ``svget_neonq`` and
``svdup_neonq``. These names are taken from ``svget/svset[2|3|4]``. and 
``svdupq`` from the SVE ACLE specifications. The ``q`` is from the NEON 64/128 bit naming convention of the NEON intrinsics.
```
Please add a link to the specs and double check their official name:
1. SVE ACLE: https://developer.arm.com/documentation/100987/
2. NEON intrinsics: https://github.com/ARM-software/acle/tree/main/neon_intrinsics

fpetrogalli (changes requested):

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
MDevereau: .rst links or github links?

fpetrogalli: rst, the syntax should be `text <URL>`_

fpetrogalli (changes requested):
Screenshot 2021-11-10 at 17 42 41

The sectioning is wrong.... The intrinsic should just be listed under 5.20? See https://github.com/MDevereau/acle/blob/next-release/main/acle.rst#id128 and https://github.com/MDevereau/acle/blob/next-release/main/acle.rst#architectural-extension-bridges

So what renders as section 1.11.1 should be moved to render as section 5.20.1

Francesco

fpetrogalli:
@pthariensflame - thank you for your review! Is it OK if I add you to the list of contributors to the project for the review? Your avatar will show up here: https://github.com/ARM-software/acle#contributors-

fpetrogalli:
@allcontributors please add @peterwaller-arm for review.

fpetrogalli:
@allcontributors please add @georges-arm  for review.

fpetrogalli:
@allcontributors please add @MDevereau  for code.

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,78 @@
...
+intended function from its parameter types, but not from the return type.
+Therefore we adopt the NEON convention of using the letter `q` in the name to
+indicate a 128-bit vector in the return type. e.g. ``svget_neonq[_u16]``. If
+64-bit vectors are added later they will omit the `q`.
+
+NEON Vector duplication across all SVE vectors
+----------------------------------------------
+
+``svdup_neonq`` is added as a convenience function to avoid casting from
+NEON to SVE as a seperate step when duplicating NEON vectors into SVE vectors.
```
fpetrogalli, line 63:
```suggestion
NEON to SVE as a separate step when duplicating NEON vectors into SVE vectors.
```

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,78 @@
+=======================================
+Design Document for ARM_NEON_SVE_BRIDGE
+=======================================
+
+This document describes the design of the intrinsics defined in
+("NEON_SVE bridge")<../acle.rst#neon-sve-bridge>. These C intrinsics enable the
```
fpetrogalli, line 6: Sorry, the linkin this line seems to be broken (see https://github.com/MDevereau/acle/blob/next-release/main/design_documents/ARM_NEON_SVE_BRIDGE.rst#design-document-for-arm_neon_sve_bridge).

I *think* you might want to replace it with the following one:
```suggestion
`NEON_SVE bridge <../acle.rst#neon-sve-bridge>`_. These C intrinsics enable the
```

Please double check it works by browsing the rendering at https://github.com/MDevereau/acle/blob/next-release/main/design_documents/ARM_NEON_SVE_BRIDGE.rst#design-document-for-arm_neon_sve_bridge

fpetrogalli (changes requested):
Apologies for asking another round, but it is really minor uncontroversial things.

Also, please resolve the conflicts with `arm-software/next-release`.

Francesco

main/acle.rst
```diff
@@ -5807,3 +5820,183 @@ Instructions
 +---------------------------------+----------------+------------+-----------------+
 
 These intrinsics are available when ``arm_acle.h`` is included.
+
+Architectural Bridges
```
fpetrogalli, line 5824: nit: can we choose a common title for the top section on bridges? I vote for  `Architectural Extension Bridges` over `Architectural Bridges`.

```suggestion
Architectural Extension Bridges
```

fpetrogalli (changes requested):

fpetrogalli:
@allcontributors please add @rsandifo-arm for review.

fpetrogalli (approved):

pthariensflame:
> @pthariensflame - thank you for your review! Is it OK if I add you to the list of contributors to the project for the review? Your avatar will show up here: https://github.com/ARM-software/acle#contributors-

Absolutely; go ahead and add me!

main/acle.rst
```diff
@@ -5822,3 +5835,183 @@ Instructions
...
+|                                                                                |
+|      svuint32_t svset_neonq[_u32](svuint32_t vec, uint32x4_t subvec)           |
++--------------------------------------------------------------------------------+
+| .. code:: c                                                                    |
+|                                                                                |
+|      svuint64_t svset_neonq[_u64](svuint64_t vec, uint64x2_t subvec)           |
++--------------------------------------------------------------------------------+
+| .. code:: c                                                                    |
+|                                                                                |
+|      svfloat16_t svset_neonq[_f16](svfloat16_t vec, float16x8_t subvec)        |
```
rsandifo-arm, line 5900: I think there should be a bfloat16 version too.  For SVE, this kind of _bf16 function is available unconditionally.

main/acle.rst
```diff
@@ -5822,3 +5835,183 @@ Instructions
...
++--------------------------------------------------------------------------------+
+| .. code:: c                                                                    |
+|                                                                                |
+|      svfloat64_t svset_neonq[_f64](svfloat64_t vec, float64x2_t subvec)        |
++--------------------------------------------------------------------------------+
+
+``svget_neonq``
+---------------
+
+These intrinsics return a NEON vector from the first 128 bit subvector of
```
rsandifo-arm, line 5914: Same comments here as for `vset_neonq`:

* we need to define what happens for big-endian targets
* there should be a bf16 version too

main/acle.rst
```diff
@@ -5822,3 +5835,183 @@ Instructions
...
+|                                                                                |
+|      svuint32_t svdup_neonq[_u32](uint32x4_t vec)                              |
++--------------------------------------------------------------------------------+
+| .. code:: c                                                                    |
+|                                                                                |
+|      svuint64_t svdup_neonq[_u64](uint64x2_t vec)                              |
++--------------------------------------------------------------------------------+
+| .. code:: c                                                                    |
+|                                                                                |
+|      svfloat16_t svdup_neonq[_f16](float16x8_t vec)                            |
```
rsandifo-arm, line 6008: bf16 here too

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,78 @@
...
+duplicating NEON vectors across SVE vectors.
+
+Location
+########
+
+The intrinsics are defined in a header file called 
+``arm_neon_sve_bridge.h``.
+
+This header file includes ``arm_neon.h`` and ``arm_sve.h``.
+It also defines ``__ARM_NEON_SVE_BRIDGE`` as 1.
```
rsandifo-arm, line 17: Would need to be updated if we do change how this macro is defined.

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,78 @@
...
+``svdupq_lane``. For example:
+
+``svuint64_t sve_vector_a = ...;``
+
+``uint64x2_t neon_vec = svget_neonq_u64(svdupq_lane_u64(sve_vector_a, 1));``
+
+Why not enable the insertion of a 128-bit NEON vector into a specific SVE subvector?
+------------------------------------------------------------------------------------
+
+This behaviour is not currently available with SVE intrinsics. Such an
```
rsandifo-arm, line 43: For me, the justification is more along the lines:

> `arm_sve.h` does not currently provide any intrinsics for inserting individual elements into a specific lane of an SVE vector.  The approach should be consistent between `arm_sve.h` and `arm_neon_sve_bridge.h`.

That is, if we _did_ provide methods for inserting both individual elements and whole NEON vectors, I think the latter _would_ belong in `arm_neon_sve_bridge.h` rather than `arm_sve.h`.  But I don't think it makes sense to provide functions for whole NEON vectors when no corresponding functions exist for individual elements.

I think the nop design justification applies here too, but it's probably not worth repeating.

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,78 @@
...
+Therefore we adopt the NEON convention of using the letter `q` in the name to
+indicate a 128-bit vector in the return type. e.g. ``svget_neonq[_u16]``. If
+64-bit vectors are added later they will omit the `q`.
+
+NEON Vector duplication across all SVE vectors
+----------------------------------------------
+
+``svdup_neonq`` is added as a convenience function to avoid casting from
+NEON to SVE as a separate step when duplicating NEON vectors into SVE vectors.
+``svdup_neonq`` also prevents initializing none NEON bits in an SVE vector
```
rsandifo-arm, line 64: s/none NEON/non-NEON/

main/acle.rst
```diff
@@ -5822,3 +5835,183 @@ Instructions
...
+
+
+The NEON_SVE Bridge adds intrinsics that allow conversions between NEON and
+SVE vectors.
+
+These intrinsics are defined in the header file ``arm_neon_sve_bridge.h``.
+
+Including this header_file will include ``arm_neon.h`` and ``arm_sve.h``.
+
+Including this header file will define ``__ARM_NEON_SVE_BRIDGE`` as 1.
```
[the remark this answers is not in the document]
fpetrogalli: Yep - good point @rsandifo-arm ! I think that we should raise an error if the header file is included but the compiler is not defining the `__ARM_NEON_SVE_BRIDGE`. The macro should be defined every time SVE is present, therefore it seems to me that it should have the same meaning of `__ARM_FEATURE_SVE`. 

This might mean that we don't need to use any feature macro, and make the bridge available unconditionally every time we include the header file and we target SVE or SVE2 with the compiler?

Given that the bridge will include `arm_sve.h`, the error will be pretty clear if we want to use the bridge in a compiler invocation that is not targeting SVE.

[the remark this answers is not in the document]
georges-arm: I guess the question is how best to allow the user to know whether `#include "arm_neon_sve_bridge.h"` will work or not, given that compilers with `arm_sve.h` and without the bridge already exist. If `__has_include` or similar is present then I agree it is possibly redundant, but I don't know if there is a precedent for relying on that?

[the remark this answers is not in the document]
rsandifo-arm: Yeah, AIUI, the feature macros are provided for the reason that @georges-arm mentioned.  E.g. the canonical way of including `arm_sve.h` is:
```
#ifdef __ARM_FEATURE_SVE
#include <arm_sve.h>
#endif
```
I'm pretty sure no-one does that, but it's the canonical way nonetheless.

The idea would be the same here.  The compiler defines `__ARM_NEON_SVE_BRIDGE` to advertise to the source code that the compiler provides an `arm_neon_sve_bridge.h`.  The code can then include the header file when available and (if it wants to) fall back to something else otherwise.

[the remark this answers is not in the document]
fpetrogalli: I see, OK. So we do the following:

1. we introduce the macro in the specs but not in the header file;
2. we require the compiler to set it to 1 when targeting SVE;
3. We raise an error if the macro is not active and the header file is included via `#ifndef __ARM_NEON_SVE_BRIDGE #error "..." #endif` in the header file.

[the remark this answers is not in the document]
rsandifo-arm: 1. 👍 
2. We're probably in agreement and I'm probably just being overly pedantic, but: compilers should only set the macro to 1 if they provide the new header file.  Old compilers that target SVE would still be conforming (they would not define the macro and would not provide the header file).
3. I don't think the error is necessary.  The macro is only there to tell the user whether a feature is available.

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
MDevereau: done

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
MDevereau: done

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
MDevereau: done.

main/acle.rst
MDevereau: done.

main/acle.rst
MDevereau: done.

main/acle.rst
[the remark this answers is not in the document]
MDevereau: done.

main/acle.rst
MDevereau: added bf16 version. big-endian isnt yet resolved.

main/acle.rst
```diff
@@ -5822,3 +5835,183 @@ Instructions
...
+
+``svset_neonq``
+---------------
+
+These intrinsics set NEON vector ``subvec`` into the first 128 bit subvector of
```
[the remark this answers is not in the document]
rsandifo-arm: Summarising our off-line discussion:

I agree with @peterwaller-arm that if we're not sure, specifying that the functions copy bit-for-bit (as a .Q) rather than lane-for-lane seems safer.  Users will then have a direct way of getting the behaviour they want, whether that's no code or a shuffle (using an additional Neon intrinsic call).  If instead we define the intrinsics to include a shuffle then users who don't want one would have to do a compensating shuffle and rely on the compiler optimising the sequence away.

If it turns out that the shuffle version is more generally useful, we could add new bridge functions that do that in future.  They would behave the same as the current ones on little-endian.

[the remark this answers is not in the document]
MDevereau: Resolving this and https://github.com/ARM-software/acle/pull/72#discussion_r751218453 as resolved then.

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,78 @@
...
+``svuint64_t sve_vector_a = ...;``
+
+``uint64x2_t neon_vec = svget_neonq_u64(svdupq_lane_u64(sve_vector_a, 1));``
+
+Why not enable the insertion of a 128-bit NEON vector into a specific SVE subvector?
+------------------------------------------------------------------------------------
+
+arm_sve.h does not currently provide any intrinsics for inserting individual
+elements into a specific lane of an SVE vector. The approach should be
+consistent between arm_sve.h and arm_neon_sve_bridge.h
```
peterwaller-arm, line 45: Nit: elsewhere these header files (across the above three lines) appear in double backticks.

main/acle.rst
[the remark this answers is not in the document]
MDevereau: Added a section to the design document detailing this decision

main/design_documents/ARM_NEON_SVE_BRIDGE.rst
```diff
@@ -0,0 +1,86 @@
...
+indicate a 128-bit vector in the return type. e.g. ``svget_neonq[_u16]``. If
+64-bit vectors are added later they will omit the `q`.
+
+Why not shuffle the vector bits for big-endian?
+-----------------------------------------------
+
+Copying vectors bit-for-bit as a Q vector rather than lane-for-lane is safer
+by default for setting, getting and duping. Users requiring lane-shuffling can
+then shuffle the bits as required. Shuffling bridge intrinsics can be added if
+enough interest is shown.
```
fpetrogalli, line 65:
```suggestion
Copying vectors bit-for-bit as a Q vector rather than lane-for-lane is safer
by default for setting, getting and duplicating lanes. Users requiring lane-shuffling can
then shuffle the bits as required. Shuffling bridge intrinsics can be added if
enough interest is shown.
```

fpetrogalli (commented):
nit!

main/acle.rst
```diff
@@ -5822,3 +5835,195 @@ Instructions
...
+
+``__ARM_NEON_SVE_BRIDGE`` is defined as 1 when ``arm_neon_sve_bridge.h`` is
+available.
+
+Including this header_file will include ``arm_neon.h`` and ``arm_sve.h``.
+
+``svset_neonq``
+---------------
+
+These intrinsics set the first 128 bits of SVE vector ``vec`` to ``subvec``.
```
peterwaller-arm, line 5861: I've had a request to mention that passing `svundef` as the `vec` parameter allows the compiler to reuse the SVE register corresponding to the NEON register. Here is some suggested language, please can someone raise their hand if they can improve it :)

```suggestion
These intrinsics set the first 128 bits of SVE vector ``vec`` to ``subvec``.
When ``svundef`` is passed as the ``vec`` parameter, the compiler is able
reuse the SVE register overlapping the NEON input without generating
additional instructions.
```

peterwaller-arm (commented):
Just noting that someone has suggested the name `sv(g|s)et_lowq` rather than `sv(g|s)et_neonq` to improve consistency with the NEON API. Such a rename is straightforward to do and seems reasonable to me. Please comment if have an opinion either way.

fpetrogalli:
> Just noting that someone has suggested the name `sv(g|s)et_lowq` rather than `sv(g|s)et_neonq` to improve consistency with the NEON API. Such a rename is straightforward to do and seems reasonable to me. Please comment if have an opinion either way.

I prefer `sv[get|set]_neonq` In Neon, low pairs with the concept of high, and together they define the concept of splitting the vector in 2 halves. `svget_lowq` could be interpreted like getting the lower half of an SVE vector, which is not the case.

main/acle.rst
fpetrogalli: Sorry, I don' understand what is the issue that this change of wording is trying to address.

peterwaller-arm: My intent was to try and provide a hint to the user how they might use this. But maybe this is not an appropriate place to do that, so I am happy to drop this suggestion. I found it difficult to shoehorn it in here with appropriate language.

rsandifo-arm, line 5861: I think this is subtle enough that it would be worth expanding.

```suggestion
These intrinsics set the first 128 bits of SVE vector ``vec`` to ``subvec``.
That is, bit *i* of the result is equal to:
* bit *i* of ``subvec`` if *i* < 128
* bit *i* of ``vec`` otherwise

.. note::

   On big-endian targets, this leaves lanes in a different
   order from the “native” SVE order.  For example, if ``subvec`` is
   ``int32x4_t``, then on big-endian targets, the first memory element
   is in lane 3 of ``subvec`` and is therefore in lane 3 of the returned
   SVE vector.  Using ``svld1`` to load elements would instead put the
   first memory element in lane 0 of the returned SVE vector.
```

rsandifo-arm: The change looks good to me FWIW.  Might be worth making it a separate paragraph though.

rsandifo-arm:
> > Just noting that someone has suggested the name `sv(g|s)et_lowq` rather than `sv(g|s)et_neonq` to improve consistency with the NEON API. Such a rename is straightforward to do and seems reasonable to me. Please comment if have an opinion either way.
> 
> I prefer `sv[get|set]_neonq` In Neon, low pairs with the concept of high, and together they define the concept of splitting the vector in 2 halves. `svget_lowq` could be interpreted like getting the lower half of an SVE vector, which is not the case.

Not a strong opinion, but +1 FWIW.  Some other reasons in favour of keeping `neon`:
* The `q`/not-`q` convention is a Neon one.  Having `neon` in the name helps to contextualise the `q`.
* The 64-bit version (if we added it) would otherwise be `svset_low`, which might be unintuitive for 32-bit and smaller elements.  E.g. `svset_low_s32` would set the low 2 s32s, rather than the low s32.
* These intrinsics are dumping a raw Neon vector (with Neon lane ordering) into the low part of an SVE vector (with SVE lane numbering).  So on big-endian, the first memory element of an s32 Neon vector becomes lane 3 of the SVE vector, rather than lane 0 of the SVE vector.

main/acle.md
```diff
@@ -1648,6 +1650,11 @@ be found in [[BA]](#BA).
 | `__ARM_SIZEOF_WCHAR_T` [[ssec-Imptype]](#implementation-defined-type-properties)                                         | Size of `wchar_t`: 2 or 4                                                                          | 2           |
 | `__ARM_WMMX` [[ssec-WMMX]](#wireless-mmx)                                                         | Wireless MMX extension (32-bit-only)                                                               | 1           |
 
+## NEON-SVE Bridge macros
+
+``__ARM_NEON_SVE_BRIDGE`` is defined to 1 if NEON-SVE Bridge
```
fpetrogalli, line 1655: Please use single quotes for fixed-font text (all through the patch, I will not mark the other places). It also seems that there is a duplicate word?

```suggestion
`__ARM_NEON_SVE_BRIDGE` is defined to 1 if
```

main/acle.md
```diff
@@ -5064,3 +5071,91 @@ transaction.
 | uint64_t __ttest (void)                       | -                | Xt -> result   | ttest <Xt>        |
 
 These intrinsics are available when `arm_acle.h` is included.
+
+# Architectural Extension Bridges
+
+## NEON-SVE Bridge
+
+The NEON_SVE Bridge adds intrinsics that allow conversions between NEON and
+SVE vectors.
```
fpetrogalli, line 5080: Nit:
```suggestion
The NEON_SVE Bridge adds intrinsics that allow conversions between NEON and SVE vectors.
```

main/acle.md
```diff
@@ -5064,3 +5071,91 @@ transaction.
...
+SVE vector.  Using ``svld1`` to load elements would instead put the
+first memory element in lane 0 of the returned SVE vector.
+
+When ``svundef`` is passed as the ``vec`` parameter, compilers are able
+to reuse the SVE register overlapping the NEON input without generating
+additional instructions.
+
+| **Instances**                                                                |
+|------------------------------------------------------------------------------|
+| ```svint8_t svset_neonq[_s8](svint8_t vec, int8x16_t subvec)```              |
```
fpetrogalli, line 5110: Single quotes here too:

```suggestion
| `svint8_t svset_neonq[_s8](svint8_t vec, int8x16_t subvec)`                  |
```

main/design_documents/ARM_NEON_SVE_BRIDGE.md
```diff
@@ -0,0 +1,69 @@
+# Design Document for ARM_NEON_SVE_BRIDGE
+
+This document describes the design of the intrinsics defined in
+`NEON_SVE Bridge <../acle.md#neon-sve-bridge>`_. These C intrinsics enable the
```
fpetrogalli, line 4: I this this link should become:

```suggestion
[NEON_SVE Bridge](../acle.md#neon-sve-bridge). These C intrinsics enable the
```

(you can see that the link is not rendered here: https://github.com/MDevereau/acle/blob/next-release/main/design_documents/ARM_NEON_SVE_BRIDGE.md)

main/design_documents/ARM_NEON_SVE_BRIDGE.md
```diff
@@ -0,0 +1,69 @@
...
+## Location
+
+The intrinsics are defined in ``arm_neon_sve_bridge.h``. If
+``arm_neon_sve_bridge.h`` is available on the compiler,
+``__ARM_NEON_SVE_BRIDGE`` is defined as 1. Users include the NEON/SVE Bridge
+as follows:
+
+```
+#ifdef __ARM_NEON_SVE_BRIDGE
+#include "arm_neon_sve_bridge.h"
```
fpetrogalli, line 17: Nit: should this be 
```suggestion
#include <arm_neon_sve_bridge.h>
```

main/design_documents/ARM_NEON_SVE_BRIDGE.md
```diff
@@ -0,0 +1,69 @@
...
+### Intrinsic Naming
+
+The intrinsics are named ``svset_neonq``, ``svget_neonq`` and
+``svdup_neonq``. These names are taken from ``svget/svset[2|3|4]`` and 
+``svdupq`` from the [SVE ACLE specification](https://developer.arm.com/documentation/100987/).
+The ``q`` is from the 64/128 bit naming convention of the 
+[NEON](https://github.com/ARM-software/acle/tree/main/neon_intrinsics/)
+intrinsics.
+
+
```
fpetrogalli, line 69: nit:
```suggestion
```

fpetrogalli (commented):
@MDevereau - thank you for the rst->md conversion.

Just a couple of nits.

Francesco

main/acle.md
MDevereau: do we not care about the 80 char line limit anymore?

fpetrogalli: Yes, we do. Ignore my comment. I got confused by a line wrap while reviewing this on a small screen. Apologies for the distraction.

fpetrogalli (approved):
LGTM! Thank you @MDevereau !

pthariensflame:
> @pthariensflame - thank you for your review! Is it OK if I add you to the list of contributors to the project for the review? Your avatar will show up here: https://github.com/ARM-software/acle#contributors-

Is this happening?  Sorry to be pushy.  😅

pthariensflame:
@fpetrogalli

PR #153, opened 2022-01-04
Prepare 2021Q4 release

fpetrogalli:
      fixing the bug yourself).
      found in the [contribution
      request.

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -205,8 +205,6 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * Added ToC declaration at the start of the document.
 * Changed section header reference links.
 * Changed the following section header texts:
- * [ssec-NEON-intrinsics-concepts](#neon-intrinsics-concepts) has changed from
-   "Concepts" to "Neon Intrinsics Concepts".
  * [ssec-bf16-scalar](#availability-of-16-bit-brain-floating-point-arithmetic-scalar-intrinsics)
    has changed from "16-bit brain floating-point arithmetic scalar intrinsics"
    to "Availability of 16-bit brain floating-point arithmetic scalar intrinsics".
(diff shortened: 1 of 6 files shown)


main/acle.md
```diff
@@ -1104,9 +1108,10 @@ This macro is defined for AArch32 only.
 ### 32-bit SIMD instructions
 
 `__ARM_FEATURE_SIMD32` is defined to 1 if the 32-bit SIMD instructions
-are supported and the intrinsics defined in [ssec-32SIMD](#simd-intrinsics-for-32-bit-operations) are
-available. This also implies support for the GE global flags which
-indicate byte-by-byte comparison results.
+are supported and the intrinsics defined in
+[ssec-32SIMD](#32-bit-simd-operations) are available. This also
```
ValeriaDaneva, line 1112: Wouldn't this section link not work when  the document is converted to pdf?

main/acle.md
```diff
@@ -2700,7 +2705,7 @@ operations [[G.191]](#G.191):
 This example assumes the implementation preserves the Q flag on return
 from an inline function.
 
-## SIMD intrinsics for 32-bit Operations
+## <span id="32-bit-simd-operations">32-bit SIMD Operations</span>
```
ValeriaDaneva, line 2708: Does using `<span>` work after pdf conversion?

morello/morello.md
```diff
@@ -159,6 +159,10 @@ All content in this document is at the **Alpha** quality level.
 
 * Update the copyright statement in section [Copyright](#copyright).
 * Add a reference to the [CHERI-HYBRID](https://github.com/CTSRD-CHERI/cheri-hybrid-c-guide) *CHERI Hybrid C/C++ Programming Guide* in section [Scope](#scope).
+* Converted document sources from reStructuredText (`.rst`) to
+  Markdown (`.md`). The [`pandoc`](https://pandoc.org/) is now used to
```
ValeriaDaneva, line 163:
```suggestion
  Markdown (`.md`). [`Pandoc`](https://pandoc.org/) is now used to
```
This suggestion applies for every instance this change is described. Alternatively, "The pandoc command" should also work.

main/acle.md
fpetrogalli: Ah - good catch - the hyperlink was taking to the entry of the section in the TOC, not in the body of the specs. I have moved the `span` right before the sectioning, now it works. You can check in the artifacts of the latest build.

main/acle.md
fpetrogalli: Nope, see my previous comment in https://github.com/ARM-software/acle/pull/153#discussion_r778448755

morello/morello.md
fpetrogalli: I opted for `The tool pandoc ...`

ValeriaDaneva (approved):

PR #172, opened 2022-02-25
[main] Add support for SVE

rsandifo-arm:
This patch incorporates the main text of the SVE ACLE, moving
the SVE ACLE from “beta” to “release” status at the same time.
The separate beta document is no longer going to be maintained.

For now the text refers to the interactive intrinsics on
developer.arm.com for the full list of intrinsics.  However,
it does include the (hand-written) table that maps SVE instructions
to their associated intrinsics, with the intrinsics again being
links to developer.arm.com.  This is often more precise than
it was in the separate beta doc, since the interactive intrinsics
tool allows the functions to be filtered by element size or data type.
However, there are some cases in which it isn't possible to narrow
the search term down exactly.  Examples include:

* Contiguous extending loads, where an explicit suffix is always
  required.  It is not possible to include those suffixes while
  excluding _gather.

* Stores, where the element size filter has no effect.
  (The filter applies to the return type, which is always
  void for stores.)

The new links should be better on balance though.

For now the instruction-to-intrinsic table includes only the
FEAT_SVE instructions.  A later patch will add the rest.

Other notes:

Quoting a table (as for the table with a “Type category” heading)
caused pandoc with a default LaTeX template to blow up with:

---------------------------------------------------------------
! pdfTeX error (ext4): \pdfendlink ended up in different nesting level than \pd
fstartlink.
\AtBegShi@Output ...ipout \box \AtBeginShipoutBox
                                                  \fi \fi
l.8921 \end{quote}

!  ==> Fatal error occurred, no output PDF file produced!
---------------------------------------------------------------

But it works (and produces correctly-formatted output) with our
template.  The patch therefore makes hyperref-check.sh use the
template too.

Things to do separately:

* Replace existing non-SVE uses of “vector” with something more
  specific, in cases where “vector” refers to Advanced SIMD vectors
  in particular.

* List other SVE instructions, not just the FEAT_SVE ones.

* Improve the formatting of level 5 headers in the PDF output.

* Much longer term: create a separate document that lists the
  SVE intrinsics, rather than relying on developer.arm.com.

---
name: Pull request

---

If this PR is about a bugfix:

guidelines](../CONTRIBUTING.md), in particular the section on [submitting
a proposal](../CONTRIBUTING.md#proposals-for-new-content).

checklist.

      fixing the bug yourself).
* [X] I have udated the `Copyright` section of the sources of the
      found in the [contribution
      request.


fpetrogalli (changes requested):
@rsandifo-arm  - this looks great, thank you for publishing it!

The mapping assembly -> intrinsic is fantastic, thank you.

I have just pointed out some minor issues for the moment, I'll have to do a deeper review.

Francesco

main/acle.md
```diff
@@ -237,6 +237,12 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * In [Data types](#data-types), clarified that `__fp16` and `__bf16` are
   predefined types whereas vector types like `int32x4_t` are not.
 * Moved the [Future directions](#future-directions) chapter to the end.
+* Added support for SVE and SVE2. This incorporates the final 00bet6
```
fpetrogalli, line 240: Maybe create a link for the text `support for SVE and SVE2` into the main section of the SVE header file`<arm_sve.h>`?

main/acle.md
```diff
@@ -4726,6 +4881,2404 @@ assertions, which may speed up access to aligned data (and will fault
...
+>
+> *   Conversely, extending loads always specify an explicit extension
+>     type, since this information is not available in the suffixes.
+>     A sign-extending load has the same base as the architectural
+>     instruction (e.g. `ld1sb`) while a zero-extending load replaces
+>     the `s` with a `u` (e.g. `ld1ub` for a zero-extending `LD1B`).
+>     Thus [`svld1ub_u32`](https://developer.arm.com/architectures/instruction-sets/intrinsics/#q=svld1ub_u32)
+>     zero-extends 8-bit data to a vector of
+>     `uint32_t`s while [`svld1sb_u32`](https://developer.arm.com/architectures/instruction-sets/intrinsics/#q=svld1sb_u32)
+>     sign-extends 8-bit data to a vector of `uint32_t`s.
```
fpetrogalli, line 6087: Why are these lines and those following quoted?

rsandifo-arm: > Why are these lines and those following quoted?

It's just a way of formatting the text to look like a description list.  There is an extension that supports description lists directly:
```
thing

:   description
```
but although pandoc supports it, github doesn't.

fpetrogalli: Why not make the description a header of level 4? This would look useful in the toc:

## SVE intrinsics
### SVE naming convention
#### base
description of base...
#### disambiguator
...
#### type0 and type1
...
#### predication
...

rsandifo-arm: It's actually a nested description list: we have another one inside the “predication” entry, so we'd still need something there.

I don't think “base” on its own would be a good section heading.  It doesn't give much context and it doesn't follow the house style of using sentence case.

peterwaller-arm (commented):
Typo. s/svbfloa16_t/svbfloat16_t/

main/acle.md
```diff
@@ -4726,6 +4881,2404 @@ assertions, which may speed up access to aligned data (and will fault
...
+Vector type & \multicolumn{9}{l}{Element selected by each \texttt{svbool\_t} bit (from lsb)}\tabularnewline
+\midrule
+\endhead
+\texttt{svint8\_t}  & 0 & 1 & 2 & 3 & 4 & 5 & 6 & 7 & 8\tabularnewline
+\texttt{svint16\_t} & 0 & - & 1 & - & 2 & - & 3 & - & 4\tabularnewline
+\texttt{svint32\_t} & 0 & - & - & - & 1 & - & - & - & 2\tabularnewline
+\texttt{svint64\_t} & 0 & - & - & - & - & - & - & - & 1\tabularnewline
+\bottomrule
+\end{longtable}
+```
```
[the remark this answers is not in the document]
rsandifo-arm: Can we try it as-is first, and consider something else if it doesn't work?  I tried both the pdf and html versions locally and it seemed to come out OK.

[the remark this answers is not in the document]
fpetrogalli: > I tried both the pdf and html versions locally and it seemed to come out OK.
The HTML version is build with Jekyll, not pandoc . Or have you actually tested Jekyll locally?

main/acle.md
fpetrogalli: Oh I see - so it is more complicated than it looks. This seems to be an extra reason for using level sectioning. If `base` and others are not good titles, maybe let's try to use something along the lines of 

```
#### Base convention
#### Type0 and Type1 conventions
#### Convention for predication
##### Zeroing - `z`
....
```

main/acle.md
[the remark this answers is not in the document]
rsandifo-arm: > > I tried both the pdf and html versions locally and it seemed to come out OK.
> 
> The HTML version is build with Jekyll, not pandoc . Or have you actually tested Jekyll locally?

I tried cut-&-pasting it into a github edit of README.md and clicking “Preview”.  It seemed to render well.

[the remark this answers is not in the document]
fpetrogalli: > I tried cut-&-pasting it into a github edit of README.md and clicking “Preview”. It seemed to render well.

Oh, that's a very good data point, because I believe GitHub uses the same Jekyll both for browsing the project and for generating Pages.

You could always activate Pages on your github.io page and browse the final effect via https://rsandifo-arm.github.io/acle/

main/acle.md
```diff
@@ -4726,6 +4881,2404 @@ assertions, which may speed up access to aligned data (and will fault
...
+
+#### List of instructions
+
+This section contains a list of all FEAT_SVE instructions. For each one it
+gives a reference to the associated ACLE intrinsic or explains why no
+such intrinsic exists.
+
+| **Instruction**                      | **Intrinsic** |
+| ------------------------------------ | ------------- |
+| ABS                                  | [`svabs`](https://developer.arm.com/architectures/instruction-sets/intrinsics/#f:@navigationhierarchiesreturnbasetype=[int]&q=svabs) |
```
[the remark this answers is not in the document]
rsandifo-arm: Hmm, not sure.  I see the log also has:
```
2022-02-26T11:11:31.1678265Z [✖] https://developer.arm.com/architectures/instruction-sets/intrinsics/#q=svcntb_pat → Status: 403
```
which looks like a fairly innocuous uri.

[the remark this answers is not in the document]
fpetrogalli: > [...] looks like a fairly innocuous uri.

Yeah, the server is refusing to load that too because it seeing all the other requests, so it doesn't matter what the rest of the URL looks like. After a certain threshold, all request are rejected.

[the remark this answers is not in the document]
rsandifo-arm: I tried fixing(?) the path in ci.yml and the latest version came back clean.  Would feel more confident in the result if it printed more than an “OK” line though :-)

.github/workflows/ci.yml
```diff
@@ -24,4 +24,4 @@ jobs:
     - uses: actions/checkout@master
     - uses: gaurav-nelson/github-action-markdown-link-check@v1
       with:
-          config-file: '.github/ci/markdown-link-check.json'
+          config-file: '.github/workflows/markdown-link-check.json'
```
fpetrogalli, line 27: This is a surprising fix...

main/acle.md
```diff
@@ -4726,6 +4881,2404 @@ assertions, which may speed up access to aligned data (and will fault
...
+    AAPCS64 calling conventions. For example, if an implementation
+    supports some form of pragma that changes `__ARM_FEATURE_SVE_BITS`
+    within a translation unit:
+
+    ``` c
+      …pragma that changes __ARM_FEATURE_SVE_BITS to 256…
+
+      typedef svfloat32_t vec256 __attribute__((arm_sve_vector_bits(256)));
+
+      …pragma that changes __ARM_FEATURE_SVE_BITS to 512…
```
fpetrogalli, line 5717:
```suggestion
      // ...pragma that changes __ARM_FEATURE_SVE_BITS to 512...
```

fpetrogalli (changes requested):
Nits! Why use a single character for `...`?

main/acle.md
rsandifo-arm: But the current “base”, “disambiguator” etc. refer back to:

```
  svbase[_disambiguator][_type0][_type1]…[_predication]
```

Rewording them loses that link.

Also, I think the current version looks better in terms of formatting.  It was a description list in the original doc, and for instance we have:
```
  **predication**
  
  > This suffix describes the inactive elements in the result of a
  > predicated operation. It can be one of the following:
  >
  > **`z`**
  >
  > > Zero predication: set all inactive elements of the result to zero.
  > …
  > Predicated loads, predicated comparisons and predicated string match
  > operations are always zeroing operations, so for brevity, the
  > corresponding intrinsics have no predication suffix.
```
where the final paragraph is part of the general “predication” entry.

main/acle.md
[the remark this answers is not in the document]
rsandifo-arm: I need complex(ish) tables for the SVE intrinsics documentation, so I looked for a way to keep the current HTML vs. LaTeX switch.  It turns out that kramdown doesn't support “```{=format}…```”, so the updated patch takes a different approach, described in the new covering message.  It seems to give good output for HTML and PDF, although TBH the HTML table doesn't look as good with the custom styling as it did with the default GitHub one.  We might want to tweak the style later (if only for this table).

.github/workflows/ci.yml
fpetrogalli: @rsandifo-arm - I think it would make sense to submit this as a generic bugfix of the CI configuration in a separate PR.

fpetrogalli (commented):
Sorry, one more tentative :)

main/acle.md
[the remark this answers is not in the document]
rsandifo-arm: I've reworked the text and table so that we can use a simpler markdown table.

main/acle.md
```diff
@@ -4726,6 +4881,2404 @@ assertions, which may speed up access to aligned data (and will fault
...
+
+    These rules only have an effect if an implementation allows the
+    register size to vary within a translation unit. The aim is to forbid
+    function definitions and function calls that have no mapping to the
+    AAPCS64 calling conventions. For example, if an implementation
+    supports some form of pragma that changes `__ARM_FEATURE_SVE_BITS`
+    within a translation unit:
+
+    ``` c
+      …pragma that changes __ARM_FEATURE_SVE_BITS to 256…
```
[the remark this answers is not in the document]
rsandifo-arm: I've replaced `…` with `...` throughout (in code and outside).  I didn't comment the lines because I wanted to make it clear that this was supposed to be a stand-in for real code.

main/acle.md
```diff
@@ -4955,6 +5111,2370 @@ assertions, which may speed up access to aligned data (and will fault
...
+correctly when the runtime register size belongs to the intersection of
+SourceVL and ObjectVL. The behavior in other cases is undefined.
+
+The compiler might impose a fixed ObjectVL set for all code or provide a
+level of user control. It might also allow ObjectVL to vary within the
+program, either within the same translation unit or between different
+translation units. Whether and how this happens is currently outside the
+scope of ACLE.
+
+Note that the combination of a singleton SourceVL and an unrestricted
```
sallyarmneale, line 5688:
```suggestion
Note: the combination of a singleton SourceVL and an unrestricted
```

main/acle.md
```diff
@@ -4955,6 +5111,2370 @@ assertions, which may speed up access to aligned data (and will fault
...
+only a single register size, perhaps to try to improve the performance
+of surrounding code. Other combinations are also possible.
+
+This section describes ACLE features related to singleton ObjectVLs.
+They are intended to help cases in which SourceVL is the same singleton.
+
+All the features are optional: a conforming ACLE implementation does not
+need to provide them.
+
+Note that the types defined in [SVE vector types](#sve-vector-types) and
```
sallyarmneale, line 5704:
```suggestion
Note: the types defined in [SVE vector types](#sve-vector-types) and
```

main/acle.md
```diff
@@ -4955,6 +5111,2370 @@ assertions, which may speed up access to aligned data (and will fault
...
+    types](#introduction-to-fixed-length-sve-types)
+    is a singleton vector register size of N bits; and
+
+*   it is possible to apply the `arm_sve_vector_bits(N)` attribute to
+    SVE vector types and SVE predicate types, as described in
+    [The `arm_sve_vector_bits` attribute](#the-arm_sve_vector_bits-attribute).
+
+If the implementation allows ObjectVL to be changed within a translation
+unit, it must ensure that the value of the macro remains accurate after
+each change. It is therefore possible for `__ARM_FEATURE_SVE_BITS` to
```
sallyarmneale, line 5729:
```suggestion
each change. Therefore, it is possible for `__ARM_FEATURE_SVE_BITS` to
```

main/acle.md
```diff
@@ -4955,6 +5111,2370 @@ assertions, which may speed up access to aligned data (and will fault
...
+have different values at different points in a translation unit.
+
+Defining the macro to zero has the same meaning as not defining it all.
+
+#### The `arm_sve_vector_bits` attribute
+
+##### Syntax and requirements
+
+An ACLE implementation may provide the following GNU-style type
```
sallyarmneale, line 5738:
```suggestion
An ACLE implementation might provide the following GNU-style type
```

main/acle.md
```diff
@@ -4955,6 +5111,2370 @@ assertions, which may speed up access to aligned data (and will fault
...
+
+  size_t size1 = sizeof(svint32_t);  // Invalid, svint32_t is sizeless
+  size_t size2 = sizeof(vec);        // OK, equals 64
+  size_t size3 = sizeof(svbool_t);   // Invalid, svbool_t is sizeless
+  size_t size4 = sizeof(pred);       // OK, equals 8
+  #endif
+```
+
+More specifically, the following rules apply to both vector and
+predicate VLST types:
```
sallyarmneale, line 5818:
```suggestion
The following rules apply to both vector and
predicate VLST types:
```

main/acle.md
```diff
@@ -4955,6 +5111,2370 @@ assertions, which may speed up access to aligned data (and will fault
...
+    *   it is an array type whose element type requires size N; or
+
+    *   it is a class or struct and:
+
+        *   at least one non-static member has a type that requires size N; or
+
+        *   a base type requires size N
+
+    Note that this definition specifically excludes reference and pointer
+    types.
```
sallyarmneale, line 5880:
```suggestion
    Note: this definition specifically excludes reference and pointer
    types.
```

main/acle.md
```diff
@@ -4955,6 +5111,2370 @@ assertions, which may speed up access to aligned data (and will fault
...
+    ``` c
+      #if __ARM_FEATURE_SVE_BITS==256 && __ARM_FEATURE_SVE_PREDICATE_OPERATORS
+      svbool_t p __attribute__((arm_sve_vector_bits(256))) = { 0, 1 };
+      #endif
+    ```
+
+    sets the first byte of `p` to 2 for both big- and little-endian targets.
+
+An implementation may provide other operators too, but it does not need
+to do so.
```
sallyarmneale, line 6161:
```suggestion
An implementation might provide other operators too, but it does not need
to do so.
```

main/acle.md
```diff
@@ -4955,6 +5111,2370 @@ assertions, which may speed up access to aligned data (and will fault
...
+
+Note that, at the time of writing, the GNU `vector_size` extension is
+not defined for `bool` elements, nor for packed vectors of single bits.
+The definition above is intended to be a conservative subset of what such
+an extension might provide.
```
sallyarmneale, line 6166:
```suggestion
Note: at the time of writing, the GNU `vector_size` extension is
not defined for `bool` elements, nor for packed vectors of single bits.
The definition above is intended to be a conservative subset of what such
an extension might provide.
```

main/acle.md
```diff
@@ -4955,6 +5111,2370 @@ assertions, which may speed up access to aligned data (and will fault
...
+> > where the `Zd` argument supplies both the values of inactive
+> > elements and the first operand to the addition. Again, the merge
+> > input does not need to be syntactically related to the result.
+>
+> **`x`**
+>
+> > “Don't-care” predication: set the inactive elements to
+> > **unknown** values. These values could be arbitrary register
+> > contents left around from a previous operation, so accidentally
+> > using the inactive elements could lead to data leakage.
```
sallyarmneale, line 6410: If this could lead to data leakage, should it be a Warning.

main/acle.md
```diff
@@ -4955,6 +5111,2370 @@ assertions, which may speed up access to aligned data (and will fault
...
+always takes a `svuint16_t` and no other form of `svclz_type_z` does.
+ACLE therefore defines shorter, overloaded, aliases that are compatible
+with both C++ overloading and C `_Generic` associations. At a
+minimum, this means that all overloaded forms of an intrinsic have the
+same number of arguments and that it is possible to select the correct
+overload by considering each argument from left to right.
+
+(Note that although the overloading scheme is compatible with `_Generic`, a C
+implementation can use some other implementation-specific way of achieving
+the same effect.)
```
sallyarmneale, line 6436:
```suggestion
(Note: although the overloading scheme is compatible with `_Generic`, a C
implementation can use other implementation-specific ways of achieving
the same effect.)
```

main/acle.md
```diff
@@ -4955,6 +5111,2370 @@ assertions, which may speed up access to aligned data (and will fault
...
+symptoms of a program going wrong rather than something that the
+programmer deliberately planted. The implementation can therefore remove
+or reorder potentially faulting operations as long as:
+
+1.  doing so would not cause a previously non-faulting program to fault;
+    and
+
+2.  faults do not move between C++ exception blocks, in cases where
+    faults are reported as exceptions.
+
```
sallyarmneale, line 6644:
```suggestion
*  doing so would not cause a previously non-faulting program to fault;
    and

*  faults do not move between C++ exception blocks, in cases where
    faults are reported as exceptions.

```

main/acle.md
```diff
@@ -4955,6 +5111,2370 @@ assertions, which may speed up access to aligned data (and will fault
...
+1.  doing so would not cause a program to raise IEEE exceptions in cases
+    where it would not previously;
+
+2.  IEEE exceptions do not move between C++ exception blocks, in cases
+    where IEEE exceptions are reported as C++ exceptions; and
+
+3.  IEEE exceptions do not move across functions that manipulate the IEEE
+    exception state.
+
```
sallyarmneale, line 6657:
```suggestion
* doing so would not cause a program to raise IEEE exceptions in cases
    where it would not previously;

*  IEEE exceptions do not move between C++ exception blocks, in cases
    where IEEE exceptions are reported as C++ exceptions; and

* IEEE exceptions do not move across functions that manipulate the IEEE
    exception state.

```

main/acle.md
```diff
@@ -4955,6 +5111,2370 @@ assertions, which may speed up access to aligned data (and will fault
...
+| *nwrite*       | the number of explicit writes to the FFR                       |
+| *lastwrite*    | the last value written to the FFR                              |
+| *nread*        | the number of explicit reads from the FFR since the last write |
+
+There are only two possible ways of modifying this FFRT:
+
+1.  Increment *nwrite* by one, set *lastwrite* to a given value, and set
+    *nread* to zero.
+
+2.  Increment *nread* by one.
```
sallyarmneale, line 6730:
```suggestion
*  Increment *nwrite* by one, set *lastwrite* to a given value, and set
    *nread* to zero.

*  Increment *nread* by one.
```

main/acle.md
```diff
@@ -4955,6 +5111,2370 @@ assertions, which may speed up access to aligned data (and will fault
...
+  ...
+  if (svptest_last(p1, p2) && x == y)
+    ...stmts...
+```
+
+Here the implementation could use PTEST to test whether the last active
+element of `p2` is active and follow it by CTERMNE to test whether
+`x` is not equal to `y`. The code should then execute `stmts` if the
+LT condition holds (i.e. if the last element is active and the
+“termination condition” `x != y` does not hold).
```
sallyarmneale, line 7426:
```suggestion
LT condition holds (that is,  if the last element is active and the
“termination condition” `x != y` does not hold).
```

sallyarmneale (changes requested):
Quite a few edit suggestions - non-technical.

main/acle.md
```diff
@@ -4955,6 +5111,2370 @@ assertions, which may speed up access to aligned data (and will fault
...
+
+This section specifies the behavior of sizeless types as an edit to the
+N1570 version of the C standard.
+
+**6.2.5 Types**
+
+> In 6.2.5/1, replace:
+>
+> > At various points within a translation unit an object type may
+> > be *incomplete* ...
```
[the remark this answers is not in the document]
rsandifo-arm: This is a quote from an external source, so I think we need to keep it as-is.

main/acle.md
rsandifo-arm: I would prefer to keep the enumeration here, since the following text refers to the “first” and “second” operation.

main/acle.md
rsandifo-arm: The meaning was more that an implementation is allowed to provide other operations too, so I changed it to “can”.

main/acle.md
```diff
@@ -4955,6 +5111,2370 @@ assertions, which may speed up access to aligned data (and will fault
...
+Any approach to defining the types would fall into one of three
+categories:
+
+1.  Limit the types in such a way that there is no concept of size.
+
+2.  Define the size of the types to be variable.
+
+3.  Define the size of the types to be constant, with the constant being
+    large enough for all possible VG or with the types pointing to
+    separate memory (as for classes like `std::string`).
```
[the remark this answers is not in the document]
rsandifo-arm: I'd prefer to keep the enumeration here, since the following discussion refers to the options by number.

main/acle.md
rsandifo-arm: I reworded this to avoid the “therefore” instead.

main/acle.md
rsandifo-arm: I went for “can choose to” instead.

main/acle.md
rsandifo-arm: I removed the “note that” instead.

sallyarmneale (approved):
Looks good to me

PR #185, opened 2022-03-29
[infra] Use `main` as the dev branck for the project.

fpetrogalli:
The `next-release` branch is going to be decommissioned because:

1. It enforces the users to check out a "non-standard" branch to check
if their bugfix applies to the current dev status of the specs. Just
reading `main` on GitHub will make things easier.
2. The allcontributos bot doesn't support the creation of PRs agains
branches other than `main`. That makes it complicated to deal with
conflicts when using allcontributors.

      fixing the bug yourself).
* [ ] I have udated the `Copyright` section of the sources of the
      found in the [contribution
      request.


CONTRIBUTING.md
```diff
@@ -72,14 +72,18 @@ to build the PDFs of the [latest official
 release](https://github.com/ARM-software/acle/releases/latest).
```
stuij, line 72: In this paragraph above, it still reads that `main is the release branch`.

CONTRIBUTING.md
```diff
@@ -72,14 +72,18 @@ to build the PDFs of the [latest official
...
-PR against the branch `next-release`.
+committed on the `main` branch. Therefore, please submit your
+PR against the branch `main`.
 
 It is good practice to track changes. In the section `Changes for next
 release`, please list any changes that you have made to the
 document. For an example, see [this
 PR](https://github.com/ARM-software/acle/pull/19).
 
+The branch `latest-release` is pointing at the more recent release of
```
stuij, line 83: I think `more recent` should be `most recent`

PR #188, opened 2022-04-01
[main] Add Alpha support for SME

rsandifo-arm:

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -8852,6 +8852,9 @@ specified in this section.
 
 ``` c
   __attribute__((arm_streaming_compatible))
+  bool __arm_has_sme(void);
+
+  __attribute__((arm_streaming_compatible))
   bool __arm_in_streaming_mode(void);
 
   // Function with external linkage.
@@ -8860,6 +8863,15 @@ specified in this section.
 ```
 
 #### Semantics
+
+**`__arm_has_sme()`**
+
+> This call returns true if the current thread “has access to SME”;
+> see [[AAPCS64]](#AAPCS64) for a more detailed definition of this term.
+>
+> One way of implementing this function is to call `__arm_sme_state`
+> and then return the top bit of X0.  See [[AAPCS64]](#AAPCS64) for
+> more details about `__arm_sme_state`.
 
 <!-- Do not remove the following `span`, they are needed to create an
 anchor that can be referred via an internal hyperlink to the paragraph
--- a/main/acle.md
+++ b/main/acle.md
@@ -8462,7 +8462,7 @@ If a function does have ZA state, the function can either share that ZA
 state with the function's caller or create new ZA state “from scratch”.
 In the latter case, it is the compiler's responsibility to free up ZA
 so that the function can use it; see the description of the lazy saving
-scheme in [[AAPCS64](#AAPCS64) for details about how the compiler
+scheme in [[AAPCS64]](#AAPCS64) for details about how the compiler
 does this.
 
 These possibilities give a one-out-of-three choice for how a function
@@ -8960,7 +8960,7 @@ The intrinsics in this section have the following properties in common:
   // slice_offset fills the role of the usual vnum parameter.
   __attribute__((arm_streaming_compatible, arm_shared_za))
   void svldr_vnum_za(uint32_t slice_base, uint64_t slice_offset,
-                     svbool_t pn, const void *ptr);
+                     const void *ptr);
 ```
 
 #### ST1B, ST1H, ST1W, ST1D, ST1Q
@@ -8998,8 +8998,7 @@ The intrinsics in this section have the following properties in common:
 ``` c
   // slice_offset fills the role of the usual vnum parameter.
   __attribute__((arm_streaming_compatible, arm_shared_za, arm_preserves_za))
-  void svstr_vnum_za(uint32_t slice_base, uint64_t slice_offset,
-                     svbool_t pn, void *ptr);
+  void svstr_vnum_za(uint32_t slice_base, uint64_t slice_offset, void *ptr);
 ```
 
 #### MOVA
--- a/main/acle.md
+++ b/main/acle.md
@@ -499,6 +499,33 @@ anchor that can be referred via an internal hyperlink to the paragraph
 > quality-of-implementation perspective, it is better to reject
 > ill-formed programs wherever possible.
 
+<!-- Do not remove the following `span`, it is needed to create an
+anchor that can be referred via an internal hyperlink to the paragraph
+2following it. --><span id="unprototyped-function"></span>
+
+**unprototyped functions**
+
+> In early versions of C, it was possible to call a function without
+> declaring it first. The function was then assumed to return an `int`.
+> For example, this was a valid complete translation unit:
+>
+> ``` c
+>   int x() { return some_func(1, 2.0, "apples"); }
+> ```
+>
+> It was also possible to declare a function's return type without specifying
+> its argument types. For example:
+>
+> ``` c
+>   double another_func();
+>   double f() { return another_func(1.0, 2, "oranges"); }
+> ```
+>
+> Functions like `some_func` and `another_func` are referred to as
+> (K&R-style) “unprototyped” functions. The first C standard categorized
+> them as an obsolescent feature and C18 removed all remaining support
+> for them.
+
 ## Conventions
 
 Most SVE ACLE intrinsics have two names: a longer unique name and a
@@ -6049,7 +6076,7 @@ The following rules apply to both vector and predicate VLST types:
         passed through “`...`”.)
 
     *   an object whose type requires some size N is passed to an
-        unprototyped function.
+        [unprototyped function](#unprototyped-function).
 
     However, the implementation does not need to diagnose these cases.
 
@@ -8597,7 +8624,7 @@ incompatible with function types that do not. For example:
 ```
 
 The function type attributes cannot be used with K&R-style
-“unprototyped” C function types. For example:
+[unprototyped function](#unprototyped-function) types. For example:
 
 ``` c
   #define ATTR __attribute__((arm_streaming))
--- a/main/acle.md
+++ b/main/acle.md
@@ -8958,7 +8958,7 @@ The intrinsics in this section have the following properties in common:
 
 ``` c
   // slice_offset fills the role of the usual vnum parameter.
-  __attribute__((arm_streaming, arm_shared_za))
+  __attribute__((arm_streaming_compatible, arm_shared_za))
   void svldr_vnum_za(uint32_t slice_base, uint64_t slice_offset,
                      svbool_t pn, const void *ptr);
 ```
@@ -8997,7 +8997,7 @@ The intrinsics in this section have the following properties in common:
 
 ``` c
   // slice_offset fills the role of the usual vnum parameter.
-  __attribute__((arm_streaming, arm_shared_za, arm_preserves_za))
+  __attribute__((arm_streaming_compatible, arm_shared_za, arm_preserves_za))
   void svstr_vnum_za(uint32_t slice_base, uint64_t slice_offset,
                      svbool_t pn, void *ptr);
 ```
@@ -9306,10 +9306,10 @@ possible to write these operations using normal C arithmetic. For example:
 #### ZERO
 
 ``` c
-  __attribute__((arm_streaming, arm_shared_za))
+  __attribute__((arm_streaming_compatible, arm_shared_za))
   void svzero_mask_za(uint64_t tile_mask);
 
-  __attribute__((arm_streaming, arm_shared_za))
+  __attribute__((arm_streaming_compatible, arm_shared_za))
   void svzero_za();
 ```
 
--- a/main/acle.md
+++ b/main/acle.md
@@ -7988,12 +7988,15 @@ following it. --><span id="non-streaming-statement"></span>
 <span id="streaming-compatible-statement"></span>
 
 ACLE provides attributes that specify whether the abstract
-machine executes statements in non-streaming mode, whether the
-abstract machine executes statements in streaming mode, or whether
-the abstract machine can execute statements in either mode. These
-statements are called “non-streaming statements”,
-“streaming statements” and “streaming-compatible statements”
-respectively.
+machine executes statements:
+
+*   in non-streaming mode, in which case they are called “non-streaming
+    statements”
+
+*   in streaming mode, in which case they are called “streaming statements”
+
+*   in either mode, in which case they are called “streaming-compatible
+    statements”
 
 At present, the classification can only be controlled at function
 granularity: the statements in a function are all non-streaming, all
@@ -8064,7 +8067,7 @@ macro to a nonzero value.
 
 In addition to [changing streaming mode locally](#changing-streaming-mode-locally),
 ACLE provides attributes for managing streaming mode across function
-boundaries. This can be useful in the situations like the following:
+boundaries. This can be useful in the following example situations:
 
 *   An SME operation is split across several cooperating subroutines
     (as is often the case). The SME operation as a whole is designed
@@ -8075,7 +8078,7 @@ boundaries. This can be useful in the situations like the following:
     Again, callers to such functions would want to avoid the
     overhead of switching modes at function call boundaries.
 
-*   Some functions might be compatible with both streaming and
+*   Some functions are compatible with both streaming and
     non-streaming mode. Marking them as “streaming-compatible”
     allows them to be called in either mode, without changing
     the vector length. For example, this could be useful for
@@ -8247,7 +8250,7 @@ following it. --><span id="non-streaming-object"></span>
     A VL-dependent object created while the abstract machine is in
     non-streaming mode is called a “non-streaming object”.
 
-If any of the following events occurs during the execution of a program
+If any of the following events occurs during the execution of a program,
 then the behavior is undefined:
 
 *   The program accesses a [non-streaming object](#non-streaming-object)
@@ -8325,7 +8328,7 @@ possible.
 
 ### Calling restrictions related to streaming mode
 
-If any of the following events occurs during the execution of a program
+If any of the following events occurs during the execution of a program,
 then the behavior is undefined:
 
 *   a [streaming caller](#streaming-caller) passes a
@@ -8426,10 +8429,13 @@ For example, PSTATE.SM is guaranteed to be 1 at the start of inline asms
 in [streaming statements](#streaming-statement) and 0 at the start of
 inline asms in [non-streaming statements](#non-streaming-statement).
 
-All inline asms must preserve the value of PSTATE.SM: an asm entered
-in streaming mode must finish in streaming mode, and an asm entered
-in non-streaming mode must finish in non-streaming mode. The behavior
-in other cases is undefined.
+All inline asms must preserve the value of PSTATE.SM.  Therefore:
+
+*   An asm entered in streaming mode must finish in streaming mode.
+
+*   An asm entered in non-streaming mode must finish in non-streaming mode.
+
+The behavior in other cases is undefined.
 
 An inline asm can temporarily switch mode internally, such as using an
 SMSTART/SMSTOP pair. However, such mode switches invalidate all Z and P
@@ -8468,7 +8474,7 @@ handles ZA:
     This is indicated by the [`arm_shared_za`](#arm_shared_za) function type
     attribute.
 
-    This case is similar in spirit to passing an uncopyable (move-only) value
+    This case is similar in concept to passing an uncopyable (move-only) value
     by reference to a C++ function:
 
     ``` c++
@@ -8511,7 +8517,7 @@ following it. --><span id="shared-za"></span><span id="private-za"></span>
 
 Reusing a term from [[AAPCS64]](#AAPCS64), the functions in category (2)
 are called “shared-ZA” functions whereas the functions in categories (1) and
-(3) are called “private-ZA” functions. In other words, “private-ZA” is the
+(3) are called “private-ZA” functions. Therefore, “private-ZA” is the
 opposite of “shared-ZA”.
 
 A program is [ill-formed](#ill-formed) if:
@@ -8640,8 +8646,7 @@ The function type attributes cannot be used with K&R-style
 
 #### `arm_streaming`
 
-This attribute applies to **function types** and specifies the following
-things:
+This attribute applies to **function types** and specifies the following:
 
 *   If the function is defined, all statements in that definition are
     [streaming statements](#streaming-statement).
@@ -8664,8 +8669,7 @@ for more information.
 
 #### `arm_streaming_compatible`
 
-This attribute applies to **function types** and specifies the following
-things:
+This attribute applies to **function types** and specifies the following:
 
 *   If the function is defined, all statements in that definition are
     by default [streaming-compatible statements](#streaming-compatible-statement).
@@ -8705,8 +8709,7 @@ for more information.
 
 #### `arm_shared_za`
 
-This attribute applies to **function types** and specifies the following
-things:
+This attribute applies to **function types** and specifies the following:
 
 *   The function has [ZA state](#za-state).
 
@@ -8721,7 +8724,7 @@ things:
 #### `arm_new_za`
 
 This attribute applies to **function definitions**. It specifies the
-following things:
+following:
 
 *   The function has [ZA state](#za-state).
 
@@ -8805,7 +8808,7 @@ a low-level feature, but it is useful for
 and could be useful for things like vector math routines.
 
 Function types with this attribute implicitly convert to function types
-that do not have the attribute. The reverse is not true. For example:
+that do not have the attribute. However, the reverse is not true. For example:
 
 ``` c
   #define ATTR __attribute__((arm_preserves_za))
@@ -8836,7 +8839,8 @@ unspecified whether the intrinsics are functions and, if so, what
 linkage they have.
 
 There are many more intrinsics than support functions, so everything
-described in this section is an intrinsic unless otherwise noted.
+described in this section is an intrinsic unless its individual
+description says otherwise.
 
 If an intrinsic is implemented as a macro rather than a function,
 the macro must behave “as if” it had the prototype and attributes
--- a/main/acle.md
+++ b/main/acle.md
@@ -296,6 +296,8 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
   No functional change intended.
 * Reordered the sections in [Change history](#change-history) in
   chronological order.
+* Added [**Alpha**](#current-status-and-anticipated-changes)
+  [support for SME](#arm_sme.h).
 
 #### Changes for next release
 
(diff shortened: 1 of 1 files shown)


georges-arm (commented):
Only managed a very brief read through, looks good. Found one small typo.

main/acle.md
```diff
@@ -4932,6 +5044,1381 @@ assertions, which may speed up access to aligned data (and will fault
...
+
+  __attribute__((arm_locally_streaming))
+  void ls_callee(svbool_t pg) {
+    // Invokes undefined behavior if called.
+  }
+
+  void n_caller(void)
+  {
+    svint8_t i = ...;
+    svbool_t b = ...:
```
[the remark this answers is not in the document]
rsandifo-arm: Thanks, fixed.

main/acle.md
```diff
@@ -4932,6 +5044,1381 @@ assertions, which may speed up access to aligned data (and will fault
...
+
+  typedef void (*n_callback_type)(void);
+  n_callback_type n_callback_ptr;
+  void n_extern_function(void);
+  void n_local_function(void) { ... }
+
+  typedef ATTR void (*s_callback_type)(void);
+  s_callback_type s_callback_ptr;
+  ATTR void s_extern_function(void);
+  ATTR void s_local_function(void) { ... }
```
fpetrogalli, line 5712: I suspect that there is an implicit meaning in the use of the `s_` and `n_` prefixes, but I have hard times finding it. Would it be possible to explain it in the text, even if just as comment in the code example?

main/acle.md
```diff
@@ -4932,6 +5044,1381 @@ assertions, which may speed up access to aligned data (and will fault
...
+    s_callback_ptr = n_extern_function;  // Ill-formed
+    s_callback_ptr = n_local_function;   // Ill-formed
+    s_callback_ptr = s_callback_ptr;     // OK
+    s_callback_ptr = s_extern_function;  // OK
+    s_callback_ptr = s_local_function;   // OK
+  }
+```
+
+The function type attributes cannot be used with K&R-style
+“unprototyped” C function types. For example:
```
fpetrogalli, line 5732: Would it make sense to add a footnote to explain or point at the definition of K&R-style
“unprototyped” C function types?

main/acle.md
```diff
@@ -4932,6 +5044,1381 @@ assertions, which may speed up access to aligned data (and will fault
...
+    *   a call to an [`arm_new_za`](#arm_new_za) function
+
+    *   a call to `setjmp`
+
+    *   a call to [`__arm_za_disable`](#__arm_za_disable)
+
+    *   a call to a non-C/C++ function that “commits a lazy save”, in the
+        sense of the [[APCS64]](#AAPCS64)
+
+    *   an inline asm that commits a lazy save
```
fpetrogalli, line 5889: Would it make sense to add a footnote that describes what a lazy save is?

main/acle.md
```diff
@@ -4932,6 +5044,1381 @@ assertions, which may speed up access to aligned data (and will fault
...
+the guarantee that is being made. The attribute does not direct the
+compiler to do anything to honor the guarantee.
+
+If a function with an `arm_preserves_za` type does not preserve ZA,
+the behavior is undefined. (There is an analogy with functions
+that are declared `noreturn` but do in fact return, and to functions
+that are declared `const` but do in fact change memory.)
+
+The attribute is intended to be useful for functions at API boundaries,
+where the compiler might not be able to “see” the definition of the
```
fpetrogalli, line 5905: Why does `see` need to be quoted?

main/acle.md
```diff
@@ -4932,6 +5044,1381 @@ assertions, which may speed up access to aligned data (and will fault
...
+function being called. As the description above implies, attaching
+`arm_preserved_za` to a [private-ZA](#private-za) function is quite
```
fpetrogalli, line 5907: What's the difference between a low level feature and a quite low level feature? :)
```suggestion
`arm_preserved_za` to a [private-ZA](#private-za) function is
```

main/acle.md
```diff
@@ -4932,6 +5044,1381 @@ assertions, which may speed up access to aligned data (and will fault
...
+    ptr2 = f1;    // OK
+    ptr2 = f2;    // OK
+  }
+```
+
+## SME functions
+
+The functions in this section are defined by the header file
+[`<arm_sme.h>`](#arm_sme.h).  Except where noted otherwise,
+the functions do not have linkage.
```
fpetrogalli, line 5938: When they do have linkage, what do you need to link them to? Is that something provided by the toolchain and transparent to the user? Or do they need to proactively link their object files against `-lWhatever`?

main/acle.md
```diff
@@ -4932,6 +5044,1381 @@ assertions, which may speed up access to aligned data (and will fault
...
+    no C or C++ type information associated with the contents of ZA.
+
+*   Intrinsic functions have a `_hor` suffix if they operate on
+    horizontal slices of a given ZA tile and a `_ver` suffix if they
+    operate on vertical slices of a given ZA tile.
+
+#### LD1B, LD1H, LD1W, LD1D, LD1Q
+
+``` c
+  // Also for _za16, _za32, _za64 and _za128.
```
fpetrogalli, line 6023: Do they have the same signature of `svld1_hor_za8`? If so, worth specifying this fact in the comment?

main/acle.md
```diff
@@ -4932,6 +5044,1381 @@ assertions, which may speed up access to aligned data (and will fault
...
+  // Return the number of doublewords in a streaming vector.
+  // Equivalent to svcntd() when called in streaming mode.
+  __attribute__((arm_streaming_compatible, arm_preserves_za))
+  uint64_t svcntsd();
+```
+
+`svcntsb()` is equivalent to an RDSVL instruction with an immediate
+operand of 1. There are no intrinsics that map directly to other
+immediate operands, or to the ADDSVL and ADDSPL instructions, but it is
+possible to write these operations using normal C arithmetic. For example:
```
fpetrogalli, line 6329:
```suggestion
`svcntsb()` is equivalent to an `RDSVL` instruction with an immediate
operand of 1. There are no intrinsics that map directly to other
immediate operands, or to the `ADDSVL` and `ADDSPL` instructions, but it is
possible to write these operations using normal C arithmetic. For example:
```

fpetrogalli (changes requested):
Hi @rsandifo-arm - mostly minor things!

Francesco

main/acle.md
rsandifo-arm: How about “rather a low-level feature”?  I'd prefer to leave some equivocation in there since to me “low-level feature” implies something that should not be used in general code.  I don't think that applies here.

main/acle.md
rsandifo-arm: A function “has linkage” if it can appear in the symbol table of an object file.  I can/should add that to the “Terms and abbreviations” section.

main/acle.md
fpetrogalli: Oh, I see, `quite a low-level feature` works well because it is follower by `..., but it is useful`. I think we can leave this one as it is. I'll mark the conversation as resolved.

main/acle.md
```diff
@@ -1426,6 +1493,21 @@ fused-multiply instructions are available in Neon also.
 intrinsics are available. This implies that `__ARM_NEON` and `__ARM_NEON_FP`
 are both nonzero.
 
+#### Scalable Matrix Extension (SME)
+
+The specification for SME is in
+[**Alpha** state](#current-status-and-anticipated-changes) and may
+change or be extended in the future.
+
+`__ARM_FEATURE_SME` is defined to 1 if the FEAT_SME instructions
```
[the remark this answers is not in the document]
rsandifo-arm: I agree it would be good to be consistent about this, but there seems to be some inconsistency in the ACLE already.  In terms of which consistency to aim for: the Arm ARM doesn't use fixed-width font for the feature names AFAICT, so would it be OK if we standardise on not using a fixed-width font in the ACLE too?

main/acle.md
rsandifo-arm: It was because it was deliberately loose language.  Compilers don't (literally) see anything.

I'll reword to use something else.

main/acle.md
rsandifo-arm: Here too I think the Arm ARM standardises on not using fixed-width font for instruction names (outside of actual code, that is).  Would it be OK to do the same here?

main/acle.md
rsandifo-arm: Yeah, I've added a note to say that these prototypes are unaffected by the type suffix.  I've also described in the MOVA section how those prototypes vary with the type suffix.

main/acle.md
```diff
@@ -4932,6 +5044,1381 @@ assertions, which may speed up access to aligned data (and will fault
...
+The following definitions are useful when describing the consequences
+of this behavior on the vector length:
+
+*   <!-- Do not remove the following `span`, it is needed to create an
+anchor that can be referred via an internal hyperlink to the section
+following it. --><span id="vl-dependent"></span>
+    A type T is said to be “VL-dependent” if an object of type T created
+    while the [abstract machine](#abstract-machine) is in streaming mode
+    has a different size from an object of type T created while the abstract
+    machine is in non-streaming mode.
```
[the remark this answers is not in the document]
rsandifo-arm: I'm not sure that's appropriate in this case.  The T isn't code, it's just a placeholder.  Non-computing articles can say things like “if company A buys company B”, and the T here has a similar role.

main/acle.md
```diff
@@ -4932,6 +5044,1381 @@ assertions, which may speed up access to aligned data (and will fault
...
+
+*   The program accesses a [non-streaming object](#non-streaming-object)
+    while the [abstract machine](#abstract-machine) is in streaming mode.
+
+*   The program accesses a [streaming object](#streaming-object) while the
+    [abstract machine](#abstract-machine) is in non-streaming mode.
+
+### Streaming callers and streaming callees
+
+If, during the execution of a program, a function F1 calls a function F2, then:
```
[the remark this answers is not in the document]
rsandifo-arm: Like the “type T” thing above, I don't think we should view F1 and F2 as code.

main/acle.md
```diff
@@ -4932,6 +5044,1381 @@ assertions, which may speed up access to aligned data (and will fault
...
+In the latter case, it is the compiler's responsibility to free up ZA
+so that the function can use it; see the description of the lazy saving
+scheme in [[AAPCS64](#AAPCS64) for details about how the compiler
+does this.
+
+There is therefore a one-out-of-three choice for how a function handles ZA:
+
+1.  The function has no [ZA state](#za-state). This is the default.
+
+2.  The function has [ZA state](#za-state) that it shares with its caller.
```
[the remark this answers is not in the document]
rsandifo-arm: I don't mind changing it if there is strong preference for bullets, but the reason for using an enumerated list here is that the list is introduced by:

> These possibilities give a one-out-of-three choice for how a function handles ZA:

I think having 1., 2. and 3. next to the three choices is more natural in that context.  In prose it would probably have been “First(ly),”, “Second(ly),” etc.

main/acle.md
```diff
@@ -4932,6 +5044,1381 @@ assertions, which may speed up access to aligned data (and will fault
...
+        shared_za_f1(shared_za);
+        ...
+      }
+    ```
+
+3.  The function has [ZA state](#za-state) that it creates “from scratch” and
+    that it does not share with its caller. This is indicated by the
+    [`arm_new_za`](#arm_new_za) function definition attribute.
+
+    This case is similar in spirit to declaring a single uncopyable C++
```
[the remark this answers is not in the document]
rsandifo-arm: This comes from the C++ standard:

> One can make a class uncopyable, i.e., move-only, by using deleted
> definitions of the copy constructor and copy assignment operator, and then
> providing defaulted definitions of the move constructor and move assignment operator.

I'll add “(move-only)”.

main/acle.md
```diff
@@ -4932,6 +5044,1381 @@ assertions, which may speed up access to aligned data (and will fault
...
+    ```
+
+<!-- Do not remove the following `span`s, they are needed to create an
+anchor that can be referred via an internal hyperlink to the section
+following it. --><span id="shared-za"></span><span id="private-za"></span>
+
+Reusing a term from [[AAPCS64]](#AAPCS64), the functions in category (2)
+are called “shared-ZA” functions whereas the functions in categories (1) and
+(3) are called “private-ZA” functions. In other words, “private-ZA” is the
+opposite of “shared-ZA”.
```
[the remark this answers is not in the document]
rsandifo-arm: The problem is that the private-ZA/shared-ZA distinction is binary whereas the choice here is one-out-of-three,

main/acle.md
```diff
@@ -4932,6 +5044,1381 @@ assertions, which may speed up access to aligned data (and will fault
...
+
+    If the inline asm does not have a `"za"` clobber but nevertheless
+    reads the current contents of ZA or changes the contents of ZA,
+    the behavior is undefined.
+
+2.  If F does not have [ZA state](#za-state), the inline asm must
+    “comply with the lazy saving scheme”, in the sense of
+    [[AAPCS64]](#AAPCS64). The behavior in other cases is undefined.
+
+    The inline asm is [ill-formed](#ill-formed) if it has a `"za"` clobber.
```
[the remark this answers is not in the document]
rsandifo-arm: The list is introduced by:

> There are then two cases:

so enumerating them felt more natural to me.

Here too I can change to a bullet list if that is strongly preferred.  I just wanted to give the reason first.

main/acle.md
rsandifo-arm: I'll add a description to the section that defines “abstract machine” etc. (with a link).

main/acle.md
```diff
@@ -4932,6 +5044,1381 @@ assertions, which may speed up access to aligned data (and will fault
...
+*   As a shorthand, an object or value is said to be VL-dependent if its
+    type is VL-dependent.
+
+    Note that an object never changes size after it has been created.
+    The “VL-dependent” classification instead decides when it is valid to
+    access the object, as described below.
+
+*   <!-- Do not remove the following `span`, it is needed to create an
+anchor that can be referred via an internal hyperlink to the section
+following it. --><span id="streaming-object"></span>
```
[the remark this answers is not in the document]
rsandifo-arm: OK, changed the affected ones to say “paragraph” instead.

main/acle.md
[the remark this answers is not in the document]
fpetrogalli: SGTM, leave them as numbered items.

main/acle.md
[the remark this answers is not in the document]
fpetrogalli: Leave it as numbered items.

main/acle.md
fpetrogalli: SGTM

fpetrogalli (approved):
Thank you @rsandifo-arm!

main/acle.md
rsandifo-arm: Now done.  (Missed this in the previous commit, sorry.)

fpetrogalli (approved):

main/acle.md
```diff
@@ -7499,6 +7653,1415 @@ This is because the ACLE intrinsic calls do not imply a particular
...
+> current ZA contents. The call might also be useful for static analysis.
+
+### SME instruction intrinsics
+
+#### Common rules
+
+The intrinsics in this section have the following properties in common:
+
+*   Every argument named `tile`, `slice_offset` or `tile_mask` must
+    be an integer constant expression.
```
bryanpkc, line 8649: I just came across this ACLE draft. What was the rationale for making it the user's responsibility to index into ZA storage explicitly, instead of creating sizeless types analogous to `svint32_t`, `svfloat64_t`, etc.? We have done [some work](https://github.com/Huawei-PTLab/llvm-project/tree/scalable-matrix) in that direction, and extended ACLE to provide types such as `smint32_t` and `smfloat64_t`. Did you consider that approach at all?

rsandifo-arm: Hi  Thanks for the pointer, this looks like really nice work.  Yeah, we did consider using explicit C/C++ objects to represent the arrays, but in the end, we thought it would be better for the low-level ACLE to maintain a more direct mapping to the instructions.  There were several reasons for this:

* It seemed likely that programmers writing code specifically for SME would be aware of the number of available tiles for a given element type, would know how the tiles are arranged, and would want to use that knowledge to hide accumulation latencies.  We therefore wanted to let the programmer specify tile numbers directly if they wanted to.  In other words, we thought SME programmers would be thinking of ZA as a whole (unlike SVE programmers, who would think of vectors as individual objects rather than as 1 out of 32 vector registers).
* Although compilers are improving all the time, there are still cases where they make bad RA decisions for vector code, especially when inputs and outputs need to be tied.  The danger with treating matrices like vectors is that the cost of any compiler mistakes would be O(VL) rather than O(1).  And it can be quite difficult to work around these issues in the source code.  This might be a particular issue when “reinterpreting” 8 64-bit tiles as 1 8-bit tile (for example), since the 8 tiles would need to be in a particular order.
* If the tiles were normal C/C++ objects, a function wouldn't be able to return a scalable matrix object (in ZA) at the same time as returning a normal object (in GPRs or FPRs) unless we provided some way of putting scalable objects into structures/classes.  That would be a good thing long term, but it seems difficult to do, since C++ has a fundamental assumption that sizeof is a constant expression.
* Complex operations might be split over several subroutines.  We thought that, in those cases, the routines would be sharing ZA as a whole, so having a single `arm_shared_za` attribute seemed more convenient than having to pass multiple tiles around.

However, the above is all about the low-level ACLE interface.  I don't think this has to be an “either/or”.  It would be useful to have higher-level features too, where the compiler does more of the work.  Also, having scalable matrix types in LLVM IR sounds like it should be a good fit for Florian's Clang matrix extensions.

bryanpkc: Hi @rsandifo-arm, thank you for the thoughtful answer. I agree with you that this doesn't have to be an "either/or"; perhaps it's even possible to implement this ACLE draft on top of our work. You have raised very good points about register allocation decision and reinterpretation; we found that we would need to teach LLVM's register allocator about sub-tiles in SME to make it do a better job, and if we were to support a reinterpretation intrinsic, it would put even more constraints on RA.

About a function's inability to return a scalable matrix object at the same time as returning a normal object, wouldn't the same limitation apply to scalable vector objects as well? Or has that problem already been solved?

rsandifo-arm: Hi @bryanpkc : No, you're right, the same restriction applies to scalable vector objects as well.  The `x2_t`, `x3_t` and `x4_t` types make it possible to return up to 4 vectors at a time (although all vectors need to be the same time, which would introduce some awkward reinterprets if someone wanted to return, say, a vector of `float`s and a vector of `uint32_t`s).  But there is currently no way of returning both a scalable vector and a scalar by value, or even a scalable vector and a scalable predicate.  It's an unfortunate restriction.

It is of course possible to return things by reference, but it would be better if that wasn't necessary.

fpetrogalli: Hi both.

@bryanpkc  - thank you for raising your concerns. Please correct if I am wrong, but my understanding is that you are happy with explanations that @rsandifo-arm provided in their answer. If that is the case, are you happy for us to proceed with merging the specs for the SME ACLE as they are in this patch? We will be open to further improvements, if needed.

@rsandifo-arm  - your answer is quite useful as it provides a justification of the current shape of the SME ACLE. Do you mind adding it to and SME-specific design document in the `design_documents` folder?

Kind regards,

Francesco

main/acle.md
```diff
@@ -7499,6 +7626,1415 @@ This is because the ACLE intrinsic calls do not imply a particular
...
+  // Only if __ARM_FEATURE_SME_I16I64 != 0
+  __attribute__((arm_streaming, arm_shared_za))
+  void svusmops_za64[_u16](uint64_t tile, svbool_t pn, svbool_t pm,
+                           svuint16_t zn, svint16_t zm);
+```
+
+#### ZERO
+
+``` c
+  __attribute__((arm_streaming, arm_shared_za))
```
sdesmalen-arm, line 9038: Should the `svzero_*` intrinsics be marked as `arm_streaming_compatible`? The instruction definitions don't require them to be executed in streaming mode (they're predicated with `CheckSMEAndZAEnabled`, not `CheckStreamingSVEAndZAEnabled`

main/acle.md
```diff
@@ -7499,6 +7653,1415 @@ This is because the ACLE intrinsic calls do not imply a particular
...
+  typedef ATTR int ft2(void);  // OK
+  ATTR int f2(void) { ... }    // OK
+```
+
+### SME attributes related to streaming mode
+
+#### `arm_streaming`
+
+This attribute applies to **function types** and specifies the following
+things:
```
sallyarmneale, line 8377:
```suggestion
This attribute applies to **function types** and specifies the following:
```

main/acle.md
```diff
@@ -7499,6 +7653,1415 @@ This is because the ACLE intrinsic calls do not imply a particular
...
+
+See [Changing streaming mode locally](#changing-streaming-mode-locally)
+for more information.
+
+### SME attributes relating to ZA
+
+#### `arm_shared_za`
+
+This attribute applies to **function types** and specifies the following
+things:
```
sallyarmneale, line 8442:
```suggestion
This attribute applies to **function types** and specifies the following:
```

main/acle.md
```diff
@@ -7499,6 +7653,1415 @@ This is because the ACLE intrinsic calls do not imply a particular
...
+    its callers.
+
+*   If the function forms part of the object code's ABI, that object code
+    function has a “shared-ZA interface”; see [[AAPCS64]](#AAPCS64)
+    for more details.
+
+#### `arm_new_za`
+
+This attribute applies to **function definitions**. It specifies the
+following things:
```
sallyarmneale, line 8457:
```suggestion
This attribute applies to **function definitions**. It specifies the
following:
```

main/acle.md
```diff
@@ -7499,6 +7653,1415 @@ This is because the ACLE intrinsic calls do not imply a particular
...
+
+[`<arm_sme.h>`](#arm_sme.h) declares various support functions and
+defines various intrinsics. The support functions have
+[external linkage](#external-linkage), like standard C functions such
+as `memcpy` do. However, as noted in [Intrinsics](#intrinsics), it is
+unspecified whether the intrinsics are functions and, if so, what
+linkage they have.
+
+There are many more intrinsics than support functions, so everything
+described in this section is an intrinsic unless otherwise noted.
```
sallyarmneale, line 8572:
```suggestion
There are many more intrinsics than support functions, so unless otherwise noted, everything
described in the following section is an intrinsic .
```

sallyarmneale (changes requested):
Some minor stylistic/formatting changes

main/acle.md
bryanpkc: @fpetrogalli Sorry for the late response. I am happy with the explanations from @rsandifo-arm and have no more comment on this point.

main/acle.md
```diff
@@ -7499,6 +7653,1415 @@ This is because the ACLE intrinsic calls do not imply a particular
...
+    streaming mode is called a “streaming object”.
+
+*   <!-- Do not remove the following `span`, it is needed to create an
+anchor that can be referred via an internal hyperlink to the paragraph
+following it. --><span id="non-streaming-object"></span>
+    A VL-dependent object created while the abstract machine is in
+    non-streaming mode is called a “non-streaming object”.
+
+If any of the following events occurs during the execution of a program
+then the behavior is undefined:
```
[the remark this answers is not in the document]
rsandifo-arm: I've added the comma before “then”, but the intention was to refer to a singular event rather than a collection of events.  The wording is trying to match the style of the C/C++ standard.

main/acle.md
rsandifo-arm: I'd prefer to keep “this section” rather than “the following section”, since the intrinsic and function documentation is in (multiple) subsections of the current one.

I've changed the “unless otherwise noted” to “unless its individual description says otherwise” to try to make it less woolly.

main/acle.md
rsandifo-arm: Oops, thanks for catching that.  I've also fixed `svldr` and `svstr`.

main/acle.md
```diff
@@ -7766,6 +7920,1419 @@ This is because the ACLE intrinsic calls do not imply a particular
...
+
+In C and C++ code, access to ZA is controlled at function granularity:
+a function either uses ZA or it does not. Another way to say this is
+that a function either “has ZA state” or it does not.
+
+If a function does have ZA state, the function can either share that ZA
+state with the function's caller or create new ZA state “from scratch”.
+In the latter case, it is the compiler's responsibility to free up ZA
+so that the function can use it; see the description of the lazy saving
+scheme in [[AAPCS64](#AAPCS64) for details about how the compiler
```
pthariensflame, line 8465:
```suggestion
scheme in [AAPCS64](#AAPCS64) for details about how the compiler
```

main/acle.md
```diff
@@ -7766,6 +7920,1419 @@ This is because the ACLE intrinsic calls do not imply a particular
...
+    of a given ZA tile and a `_ver` suffix if they operate on vertical
+    slices of a given ZA tile.
+
+#### LD1B, LD1H, LD1W, LD1D, LD1Q
+
+``` c
+  // Also for _za16, _za32, _za64 and _za128 (with the same prototype).
+  __attribute__((arm_streaming, arm_shared_za))
+  void svld1_hor_za8(uint64_t tile, uint32_t slice_base,
+                     uint64_t slice_offset, svbool_t pg, const void *ptr);
```
sagarkulkarni19, line 8935: I am curious why we need the `uint64_t tile` parameter for the `_za8` version and why we need the `uint64_t slice_offset` for `_za128` version? Is it valid for either cases to take a value other than `0`?

main/acle.md
```diff
@@ -7766,6 +7920,1419 @@ This is because the ACLE intrinsic calls do not imply a particular
...
+```
+
+#### LDR
+
+``` c
+  // slice_offset fills the role of the usual vnum parameter.
+  __attribute__((arm_streaming_compatible, arm_shared_za))
+  void svldr_vnum_za(uint32_t slice_base, uint64_t slice_offset,
+                     svbool_t pn, const void *ptr);
+```
```
sagarkulkarni19, line 8964: The `LDR` instruction in the ISA is unpredicated. I am curious why this ACLE intrinsic uses a predicate?

main/acle.md
rsandifo-arm: Thanks for the catch.  I was going for `[[]]` (to be printed as `[AAPCS64]`) since that seemed to be the style preferred elsewhere for external references.

main/acle.md
rsandifo-arm: Gah, this a pasto from the LD1 functions.  Thanks for the catch.

main/acle.md
rsandifo-arm: Gah, this a pasto from the LD1 functions. Thanks for the catch.

main/acle.md
fpetrogalli: @bryanpkc - thank you for the comments. Are you happy for me to add you as a contributor to the [list of all-contributors](https://github.com/ARM-software/acle/#contributors-) for your review?

main/acle.md
fpetrogalli: @allcontributors please add @sdesmalen-arm for review.

main/acle.md
fpetrogalli: @pthariensflame  - thank you for the comments. Are you happy for me to add you as a contributor to the [list of all-contributors](https://github.com/ARM-software/acle/#contributors-) for your review?

main/acle.md
fpetrogalli: @all-contributors please add @sdesmalen-arm for review.

fpetrogalli:
@all-contributors please add @sdesmalen-arm for review.

fpetrogalli:
@sagarkulkarni19, @pthariensflame , @bryanpkc  - thank you for the comments. Are you happy for me to add you as a contributor to the [list of all-contributors](https://github.com/ARM-software/acle/#contributors-) for your review?

bryanpkc:
> Are you happy for me to add you as a contributor to the [list of all-contributors](https://github.com/ARM-software/acle/#contributors-) for your review?

Sure.

sagarkulkarni19:
> @sagarkulkarni19, @pthariensflame , @bryanpkc - thank you for the comments. Are you happy for me to add you as a contributor to the [list of all-contributors](https://github.com/ARM-software/acle/#contributors-) for your review?

Sure, sounds good.

fpetrogalli:
@all-contributors please add @bryanpkc @pthariensflame and @sagarkulkarni19 for review.

fpetrogalli:
@all-contributors please add @pthariensflame for review.

fpetrogalli:
@all-contributors please add @sagarkulkarni19  for review.

sallyarmneale:
Looks good to me.

sallyarmneale (approved):

rsandifo-arm:
I've just updated the [AAPCS64 PR](https://github.com/ARM-software/abi-aa/pull/123) to change the way that streaming-compatible functions are handled.  Rather than have a special parameter that contains PSTATE.SM, the proposal is instead to have a utility function called `__arm_sme_state` that returns the current state.  This function also returns whether the thread has access to SME and TPIDR2_EL0.

It doesn't seem appropriate to make `__arm_sme_state` directly callable from C, since the values of PSTATE.ZA and TPIDR2_EL0 are handled by the compiler and do not have values that C/C++ code could rely on.  Also, as far as the ACLE is concerned, PSTATE.SM is a property of the abstract machine, whereas `__arm_sme_state` would directly return the underlying system register.  `__arm_in_streaming_mode` already provides the C/C++ view of PSTATE.SM.

However, the information about whether the thread has access to SME should be useful to C/C++ code, so I've added an `__arm_has_sme` intrinsic for that part.

main/acle.md
```diff
@@ -7766,6 +7893,1415 @@ This is because the ACLE intrinsic calls do not imply a particular
...
+    on return from the function. That is, the function does not use ZA
+    to receive data from callers or to pass data back to callers.
+
+This attribute does not change a function's binary interface. If the
+function forms part of the object code's ABI, that object code function
+has a “private-ZA interface”, just like all other non-`arm_shared_za`
+functions do. See [[AAPCS64]](#AAPCS64) for more details about
+private-ZA interfaces.
+
+A function definition with this attribute is [ill-formed](#ill-formed)
```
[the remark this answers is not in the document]
rsandifo-arm: Yes, a call from one function with an `arm_new_za` attribute to another function with an `arm_new_za` attribute is fine.  Using `arm_new_za` is an internal implementation choice than doesn't affect the function's ABI: the attribute simply indicates that the function uses ZA to store state (and that that state isn't shared with callers).

When a function F has an `arm_new_za` attribute, the compiler must make F commit any uncommitted lazy save before F stores new data into ZA.

main/acle.md
```diff
@@ -430,12 +432,100 @@ This document uses the following terms and abbreviations.
...
+>
+> It was also possible to declare a function's return type without specifying
+> its argument types. For example:
+>
+> ``` c
+>   double another_func();
+>   double f() { return another_func(1.0, 2, "oranges"); }
+> ```
+>
+> Functions like `some_func` and `another_func` are referred to as
```
sallyarmneale, line 524: like -> such as

main/acle.md
```diff
@@ -430,12 +432,100 @@ This document uses the following terms and abbreviations.
...
+> (K&R-style) “unprototyped” functions. The first C standard categorized
+> them as an obsolescent feature and C18 removed all remaining support
```
sallyarmneale, line 526: them -> these functions

main/acle.md
```diff
@@ -7766,6 +7920,1430 @@ This is because the ACLE intrinsic calls do not imply a particular
...
+ACLE provides attributes for managing streaming mode across function
+boundaries. This can be useful in the following example situations:
+
+*   An SME operation is split across several cooperating subroutines
+    (as is often the case). The SME operation as a whole is designed
+    to execute exclusively in streaming mode, so mid-operation mode
+    switches should be avoided for performance reasons.
+
+*   A function provides a public API that is specific to SME.
+    Again, callers to such functions would want to avoid the
```
sallyarmneale, line 8078: change to should?

sallyarmneale (commented):
3 very minor suggestions

vhscampos:
Latest comments addressed in #233

PR #198, opened 2022-05-12
[main] Add instructions for SVE extensions

rsandifo-arm:
The cover note for the original SVE patch said:

    For now the instruction-to-intrinsic table includes only the
    FEAT_SVE instructions.  A later patch will add the rest.

This patch does that.  It also tightens some existing intrinsic
links to exclude unrelated functions.  For example, linking to
svcvt_ excludes svcvtlb and svcvtlt.

The patch doesn't add an entry to the changelog since it is
really just completing the existing SVE work, which is already
listed as a change in the next release.

---
name: Pull request

---

If this PR is about a bugfix:

guidelines](../CONTRIBUTING.md), in particular the section on [submitting
a proposal](../CONTRIBUTING.md#proposals-for-new-content).

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -5577,7 +5577,7 @@ ACLE provides two sets of intrinsics for converting between vector types:
     intrinsics simply reinterpret a vector of one type as a vector of another
     type, without changing any of the bits.
 
-*   The [`svcvt`](https://developer.arm.com/architectures/instruction-sets/intrinsics/#q=svcvt)
+*   The [`svcvt`](https://developer.arm.com/architectures/instruction-sets/intrinsics/#q=svcvt_)
     intrinsics instead perform numerical conversion from one type to another,
     such as converting integers to floating-point values.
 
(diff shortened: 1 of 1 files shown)


main/acle.md
```diff
@@ -6816,19 +6816,29 @@ operation that they perform.
 
 #### List of instructions
 
-This section contains a list of all FEAT_SVE instructions. For each one it
+This section contains a list of all SVE instructions. For each one it
```
fpetrogalli, line 6819: Hi @rsandifo-arm - would it make sense to group the instructions by `__ARM_FEATURE_*`?

rsandifo-arm: I guess both ways lose something.  The current way doesn't separate instructions from different ISA features, whereas grouping them based on ISA feature would lose the simple alphabetical list.

FWIW, here are some arguments in favour of keeping it as it is:

* It matches the style used in the [ISA documentation](https://developer.arm.com/documentation/ddi0602/2022-03/SVE-Instructions/), which makes it easier to do a 1:1 comparison.

* The difference between SVE and SVE2 is likely to become less important over time, at least for new authors, so having them in a single list should gradually become more useful than having separate lists.

* The requirements for some instructions, such as the F64 matmul instructions, are more complicated than a single FEAT, so even grouping by `__ARM_FEATURE_*` will miss some information.  The current approach ensures that any information about availability is provided by the ISA documentation instead (at the cost of two hops: one from the ACLE to the intrinsics, and one from the intrinsic documentation to the instruction documentation).

Also, I don't think this list is a good starting point for exploring the ISA.  The interactive intrinsics themselves are probably better for that.  I'd imagined that this list would be more useful for people who already know which instruction they want to use and want to find the associated intrinsic.

fpetrogalli: Thank you for the clear explanation! I agree with you.

fpetrogalli (changes requested):
Please add the changes to the `Changes for next release` section.

fpetrogalli (approved):
Sorry, my bad. These changes are already listed for the next release...

PR #199, opened 2022-06-16
[main] Add feature macro __ARM_FEATURE_RCPC

minglotus-6:
To guard the usage of Load-AcquirePC instructions in inline assembly, https://reviews.llvm.org/D127798 introduces macro __ARM_FEATURE_RCPC in Clang/LLVM.

This patch documents the existence of the feature MACRO, its availability and use case.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.
* [x]  When modifying content and/or its rendering, I have checked the

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -3,7 +3,7 @@ title: Arm C Language Extensions
 version: 2022Q1
 date-of-issue: 06 April 2022
 # LaTeX specific variables
-copyright-text: Copyright 2011-2022 Arm Limited and/or its affiliates <open-source-office@arm.com>.
+copyright-text: "Copyright: see section \\texorpdfstring{\\nameref{copyright}}{Copyright}."
 draftversion: true
 # Jekyll specific variables
 header_counter: true
@@ -12,6 +12,7 @@ toc: true
 
 <!--
 SPDX-FileCopyrightText: Copyright 2011-2022 Arm Limited and/or its affiliates <open-source-office@arm.com>
+SPDX-FileCopyrightText: Copyright 2022 Google LLC.
 CC-BY-SA-4.0 AND Apache-Patent-License
 See LICENSE.md file for details
 -->
@@ -113,7 +114,8 @@ about Arm’s trademarks.
 
 ## Copyright
 
-Copyright 2011-2022 Arm Limited and/or its affiliates <open-source-office@arm.com>.
+* Copyright 2011-2022 Arm Limited and/or its affiliates <open-source-office@arm.com>.
+* Copyright 2022 Google LLC.
 
 ## About this document
 
@@ -309,6 +311,7 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * Fixes for [Function Multi Versioning](#function-multi-versioning):
   * typo in `FEAT_DPB2`.
   * added `FEAT_LS64*`.
+* Added feature detection macro `__ARM_FEATURE_RCPC` for RCpc (Release Consistent processor consistent) model at [RCpc](#rcpc).
 
 ### References
 
@@ -1391,6 +1394,25 @@ execution state. Intrinsics for the use of these instructions are
 specified in [memcpy family of operations intrinsics -
 MOPS](#memcpy-family-of-operations-intrinsics---mops)
 
+### RCPC
+
+`__ARM_FEATURE_RCPC` is set if the weaker RCpc (Release Consistent
+processor consistent) model is supported. It is undefined otherwise.
+The value indicates the set of Load-Acquire and Store-Release
+instructions available. The intention is to allow programmers to guard
+the usage of these instructions in inline assembly.
+
+| **Value** | **Feature** | **Instructions**                 | **Availability**          |
+| --------- | ----------- |--------------------------------  | ------------------------- |
+| 1         | FEAT_LRCPC  |  LDAPR* instructions             | Armv8.4, optional Armv8.2 |
+| 2         | FEAT_LRCPC2 |  LDAPUR* and STLUR* instructions | Armv8.3, optional Armv8.2 |
+
+If defined, the value of `__ARM_FEATURE_RCPC` remains consistent with the decimal
+value of `LRCPC` field (bits [23:20]) in the `ID_AA64ISAR1_EL1` register.
+
+The `__ARM_FEATURE_RCPC` macro can only be implemented in the AArch64
+execution state.
+
 ## Floating-point and vector hardware
 
 ### Hardware floating point
@@ -1971,6 +1993,7 @@ be found in [[BA]](#BA).
 | [`__ARM_FEATURE_PAC_DEFAULT`](#pointer-authentication)                                                                                                  | Pointer authentication                                                                             | 0x5         |
 | [`__ARM_FEATURE_QBIT`](#q-saturation-flag)                                                                                                              | Q (saturation) flag (32-bit-only)                                                                  | 1           |
 | [`__ARM_FEATURE_QRDMX`](#rounding-doubling-multiplies)                                                                                                  | SQRDMLxH instructions and associated intrinsics availability                                       | 1           |
+| [`__ARM_FEATURE_RCPC`](#rcpc)                                                                                                                           | Release Consistent processor consistent Model (64-bit-only)                      | 1           |
 | [`__ARM_FEATURE_RNG`](#random-number-generation-extension)                                                                                              | Random Number Generation Extension (Armv8.5-A)                                                     | 1           |
 | [`__ARM_FEATURE_SAT`](#saturation-instructions)                                                                                                         | Width-specified saturation instructions (32-bit-only)                                              | 1           |
 | [`__ARM_FEATURE_SHA2`](#sha2-extension)                                                                                                                 | SHA2 Crypto extension (Arm v8-A)                                                                   | 1           |


fpetrogalli:
Hello @minglotus-6, thank you for your patch. May I ask you to go through the checklist I have added to the description of the PR? The *Contribution Agreement* in the [CONTRIBUTING](https://arm-software.github.io/acle/CONTRIBUTING.html) allows you or the company you work for to retain copyright for the changes.

Kind regards,

Francesco

lenary:
Broadly, I am supportive of this from a technical perspective. However, I do have some minor comments.

The armv8a Arm ARM (https://developer.arm.com/documentation/ddi0487/ha/) defines the following values for the ID field for LRCPC in `ID_AA64ISAR1_EL1`:
- `0b0000` corresponds to none of `FEAT_LRCPC*`
- `0b0001` corresponds to `FEAT_LRCPC` (`LDAPR*` instructions)
- `0b0010` corresponds to `FEAT_LRCPC2` (`LDAPUR*` and `STLUR*`)

It would make sense for this macro, therefore, to match these values, as this gives a better idea of which instructions these enable in inline assembly. We do something similar with `__ARM_ARCH` (though with more complex rules). I'm not sure what to do about the 0-value case: as it seems reasonable also to leave the macro undefined when you have no LRCPC features.

I would also like the description to reference the official `FEAT_*` names somewhere, for the avoidance of doubt about exactly which values correspond to which instructions.

I think leaving the macro name as `__ARM_FEATURE_RCPC` is fine, as the compiler switch to enable `FEAT_LRCPC` instructions is `+rcpc` (Documented for GNU here: https://developer.arm.com/Tools%20and%20Software/GNU%20Toolchain#Supported-Devices, but Clang/LLVM matches this), and the consistency model in the Arm ARM is called "RCpc".

minglotus-6:
> Hello @minglotus-6, thank you for your patch. May I ask you to go through the checklist I have added to the description of the PR? The _Contribution Agreement_ in the [CONTRIBUTING](https://arm-software.github.io/acle/CONTRIBUTING.html) page allows you or the company you work for to retain copyright for the changes.
> 
> Kind regards,
> 
> Francesco

Sure. I saw the checklist but was not sure if I should list them before. Did it this time.

minglotus-6:
> Broadly, I am supportive of this from a technical perspective. However, I do have some minor comments.
> 
> The armv8a Arm ARM (https://developer.arm.com/documentation/ddi0487/ha/) defines the following values for the ID field for LRCPC in `ID_AA64ISAR1_EL1`:
> 
> * `0b0000` corresponds to none of `FEAT_LRCPC*`
> * `0b0001` corresponds to `FEAT_LRCPC` (`LDAPR*` instructions)
> * `0b0010` corresponds to `FEAT_LRCPC2` (`LDAPUR*` and `STLUR*`)
> 
> It would make sense for this macro, therefore, to match these values, as this gives a better idea of which instructions these enable in inline assembly. We do something similar with `__ARM_ARCH` (though with more complex rules). I'm not sure what to do about the 0-value case: as it seems reasonable also to leave the macro undefined when you have no LRCPC features.
> 
> I would also like the description to reference the official `FEAT_*` names somewhere, for the avoidance of doubt about exactly which values correspond to which instructions.

Keeping macro value consistent with register field value makes a lot of sense to me. This allows inline assembly users to further gate the usage of `LDAPUR*` or `STLUR*` instructions. Use a table to describe the value of macro, and include `FEAT_*` in feature columns. 

I'd like to mention one thing and hear feedback :) I wonder if using bitmaps is a recommended approach, say more values could be added to `LRCPC` field of `ID_AA64ISAR1_EL1` in the future.

To elaborate,
1) The macro value is NOT described as bitmap in the current PR, imagining users can write code like

```
#if DEFINED(__ARM_FEATURE_RCPC)
  #if __ARM_FEATURE_RCPC == 2
    Use LDAPUR or STLUR instructions
  #elif __ARM_FEATURE_RCPC == 1
    Use LDAPR instructions
  #else
    Use non-RCPC instructions
#endif
```

2) On the other hand, many existing feature macros (e.g., `__ARM_FP` [1] or `__ARM_FEATURE_CDE_COPROC` [2]) is a bitmap, and valid values (`or` results of one-bit values) are documented. 

[1] https://github.com/ARM-software/acle/blob/main/main/acle.md#hardware-floating-point
[2] https://github.com/ARM-software/acle/blob/main/main/acle.md#custom-datapath-extension

> 
> I think leaving the macro name as `__ARM_FEATURE_RCPC` is fine, as the compiler switch to enable `FEAT_LRCPC` instructions is `+rcpc` (Documented for GNU here: https://developer.arm.com/Tools%20and%20Software/GNU%20Toolchain#Supported-Devices, but Clang/LLVM matches this), and the consistency model in the Arm ARM is called "RCpc".

minglotus-6:
A side-topic that's more relevant with Clang/LLVM but less relevant here, is that `--march=<other-features>+rcpc2` is not supported by Clang yet.

As a result, for armv8.2 with rcpc2 extensions, there isn't a way for user to tell compiler to set `__ARM_FEATURE_RCPC` to 2 in the current state -> compiler could set the macro to 1 if compiler user specifies `--march=+rcpc`. Yet this could be feasible to fix in Clang/LLVM.

[1] https://github.com/llvm/llvm-project/blob/40d2ef841b68f6b493ce88bd750a92105a2b567d/llvm/include/llvm/Support/AArch64TargetParser.def#L129 is where features are defined, and https://godbolt.org/z/Gh4safcx7 exemplifies it

fpetrogalli:
@minglotus-6 - thank you for updating the list. 

May I ask you to add also the copyright statement (it is required in the [contribution agreement](https://arm-software.github.io/acle/CONTRIBUTING.html)) and update the variable `draftversion` of the YAML header to `true`?

WRT the copyright statements, what you need to do is described in the checklist of the description of the PR.

Please let me know if anything is unclear!

Thank you,

Francesco

lenary (commented):
> I'd like to mention one thing and hear feedback :) I wonder if using bitmaps is a recommended approach, say more values could be added to LRCPC field of ID_AA64ISAR1_EL1 in the future.

Please do not propose a bitmap in this case. FEAT_LRCPC2 implies FEAT_LRCPC, while they have what look like orthogonal bitmap values, this is just a coincidence from assigning 1 then 2.

I much prefer your example (1), especially so programmers can write `__ARM_FEATURE_RCPC >= 1` if they only have a version which requires `LDAPR`.

With respect to the FP and CDE extensions, the values in there are orthogonal (though the FP one tells you the exact values that preprocessor definition will ever have, showing it's not fully orthogonal).

Below, I'm broadly happy, though I have one comment about the table entry.

minglotus-6:
> @minglotus-6 - thank you for updating the list.
> 
> May I ask you to add also the copyright statement (it is required in the [contribution agreement](https://arm-software.github.io/acle/CONTRIBUTING.html)) and update the variable `draftversion` of the YAML header to `true`?
> 
> WRT the copyright statements, what you need to do is described in the checklist of the description of the PR.
> 
> Please let me know if anything is unclear!
> 
> Thank you,
> 
> Francesco

Thanks for pointing this out Francesco! I've just read some guidelines inside the company I worked for and will also consult to see what I plan to fill in looks good. I'll update once I get some guidance inside the company. As you may see I'm not very familiar with this contributing process and I appreciate your patience and guidance here!

minglotus-6:
> > I'd like to mention one thing and hear feedback :) I wonder if using bitmaps is a recommended approach, say more values could be added to LRCPC field of ID_AA64ISAR1_EL1 in the future.
> 
> Please do not propose a bitmap in this case. FEAT_LRCPC2 implies FEAT_LRCPC, while they have what look like orthogonal bitmap values, this is just a coincidence from assigning 1 then 2.

Acknowledged, thanks for taking a look!
> 
> I much prefer your example (1), especially so programmers can write `__ARM_FEATURE_RCPC >= 1` if they only have a version which requires `LDAPR`.
> 
> With respect to the FP and CDE extensions, the values in there are orthogonal (though the FP one tells you the exact values that preprocessor definition will ever have, showing it's not fully orthogonal).

This makes sense.

> 
> Below, I'm broadly happy, though I have one comment about the table entry.

Applied the suggestions.

rearnsha:
Historically we have not added #defines to ACLE unless they guard, or describe, _features or intrinsics provided by ACLE_ and that are visible directly to C programmers (ie, providing macros to conditionalize assembly language is out of scope).  So exactly what does this new macro guard?

minglotus-6:
> > @minglotus-6 - thank you for updating the list.
> > May I ask you to add also the copyright statement (it is required in the [contribution agreement](https://arm-software.github.io/acle/CONTRIBUTING.html)) and update the variable `draftversion` of the YAML header to `true`?
> > WRT the copyright statements, what you need to do is described in the checklist of the description of the PR.
> > Please let me know if anything is unclear!
> > Thank you,
> > Francesco
> 
> Thanks for pointing this out Francesco! I've just read some guidelines inside the company I worked for and will also consult to see what I plan to fill in looks good. I'll update once I get some guidance inside the company. As you may see I'm not very familiar with this contributing process and I appreciate your patience and guidance here!

Hi Francesco, 
    I updated both the checklist and the PR itself after getting replies from relevant teams inside the company. Could you take a look and see if it looks good? In particular, shall I ask for an email and append it in the copyright notices? Thanks!

minglotus-6:
> Historically we have not added #defines to ACLE unless they guard, or describe, _features or intrinsics provided by ACLE_ and that are visible directly to C programmers (ie, providing macros to conditionalize assembly language is out of scope). So exactly what does this new macro guard?

This is to guard the usage of FEAT_RCPC and FEAT_RCPC2 instructions in the inline assembly. An example usage is in https://github.com/ARM-software/acle/pull/199#issuecomment-1172834348. I wonder if the use case looks reasonable, or are there better alternatives?

minglotus-6:
A bump of this PR :-)

lenary:
I am happy with this from the technical side. I find the sentence about `ID_AA64ISAR1_EL1` clear, though I realise this is not exactly how we've defined other system registers.

rsandifo-arm:
LGTM too FWIW.

Regarding the issue that @rearnsha raised: it turned out that there were already other macros that broke the original principle.  We agreed to relax the principle to something else, with a TODO to go back and make the existing features consistent.  I've filed https://github.com/ARM-software/acle/issues/215 for that.

vhscampos:
@all-contributors please add @lenary for review.

sallyarmneale (commented):
Looks good to me

PR #260, opened 2023-05-31
GCS ACLE

nsz-arm:
---
name: Pull request

---

If this PR is about a bugfix:

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.

### DIFF
--- a/main/design_documents/gcs.md
+++ b/main/design_documents/gcs.md
@@ -1,0 +1,68 @@
+# Design Document for GCS
+
+## Feature test
+
+GCS support has three levels:
+
+(1) Code generation is GCS compatible. (Compile time decision.)
+
+(2) HW supports GCS instructions. (Might be known at compile time,
+    but this is a runtime feature.)
+
+(3) GCS is enabled at runtime. (Only known at runtime.)
+
+Where (3) implies (1) and (2). In principle a user may decide to
+enable GCS even if (1) was false at compile time, but this is
+a user error. The runtime system is responsible for enabling GCS
+when (1) and (2) holds and GCS protection was requested for the
+program.
+
+(1) and (2) need feature test macros since they can be known at
+compile time.
+
+(3) can be detected using `__chkfeat(_CHKFEAT_GCS)` which is
+available without GCS support.
+
+## Intrinsics
+
+Alternative designs for the support levels at which the intrinsics
+are well defined:
+
+(A) require (3),
+
+(B) require (1) and (2) but not (3),
+
+(C) require (2) only.
+
+Simplest is (A), but it does not allow asynchronously disabling GCS,
+for that at least (B) is needed since the intrinsics must do something
+reasonable if GCS is disabled. Asynchronous disable is e.g. needed to
+allow disabling GCS at dlopen time in a multi-threaded process when
+the loaded module is not GCS compatible.
+
+(C) is similar to (B) but allows using the intrinsics even if GCS is
+guaranteed to be disabled. The intrinsics are expected to be used
+behind runtime check for (3) since they don't do anything useful
+otherwise and thus (1) and (2) are true when the intrinsics are used
+either way. With (B) it is possible to only expose the intrinsics
+at compile time if (1) is true which can be feature tested. With (C)
+there is no obvious feature test for the presence of the intrinsics.
+
+The future direction is to make intrinsics available unconditionally
+and rely on runtime checks (e.g. via function multi-versioning). So
+it makes sense to go with (C), have separate semantics defined for
+the enabled and disabled case and let user code deal with the runtime
+checks.
+
+The type of the intrinsics is based on `const void *` GCS pointer
+type and `uint64_t` GCS entry type. The GCS pointer could be
+`const uint64_t *`, but void is more general in that it allows
+different access to the GCS (e.g. accessing entries as pointers or
+bytes). A GCS entry is usually a code pointer, but the architecture
+requires it to be 8 bytes (even with ILP32) and it may be a special
+token that requires bit operations to detect, so fixed width
+unsigned int type is the most appropriate.
+
+The const qualifier is justified for GCS even if GCS stores are
+enabled because normal stores cannot modify the GCS only specific
+instructions can.


main/acle.md
```diff
@@ -4478,6 +4501,56 @@ two pointers, ignoring the tags.
...
+Reads and returns the last entry on the GCS of the current thread and
+updates the GCS pointer to point to the previous entry.  If GCS
+protection is disabled then it has no side effect and returns `NULL`.
+
+``` c
+  const void *__gcsss(const void *);
+```
+
+Switches the GCS of the current thread, where the argument is the new
+GCS pointer, and returns the old GCS pointer.  If GCS protection is
```
DanielKristofKiss, line 4551: superfluous space before `If GCS`

DanielKristofKiss (commented):
LGTM with small comments.

main/acle.md
```diff
@@ -4478,6 +4501,56 @@ two pointers, ignoring the tags.
...
+When GCS protection is enabled then function calls save the return
+address to a separate stack, the GCS, that is checked against the actual
+return address when the function returns. At runtime GCS protection can
+be disabled and then calls and returns do not access the GCS. The GCS
+grows down and a GCS pointer points to the last entry of the GCS.
+Each thread has a separate GCS and GCS pointer.
+
+To use the intrinsics, `arm_acle.h` needs to be included.
+
+The intrinsics are available if `__ARM_FEATURE_GCS_DEFAULT` is defined,
```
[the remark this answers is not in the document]
nsz-arm: i'm not against making intrinsics always available, but for GCS the intrinsics are only useful if GCS is enabled, and that can only be enabled if everything is compiled in GCS compatible mode, so there is not really any context when `__ARM_FEATURE_GCS_DEFAULT` is undef but you want the intrinsics anyway.

main/acle.md
```diff
@@ -2937,6 +2948,18 @@ inclusive. See implementation documentation for the effect (if any) of
 this instruction and the meaning of the argument. This is available only
 when compiling for AArch32.
 
+``` c
+  uint64_t __chkfeat(uint64_t);
```
john-brawn-arm, line 2952: The behaviour of this intrinsic is opposite that of the instruction (chkfeat instruction clears the bit if the feature is enabled), which I think is unique among intrinsics. It would be worthwhile specifically noting this.

Also "feature is available" should be "feature is enabled" (as the chkfeat instruction will leave the bit alone if the feature is available but disabled).

main/acle.md
```diff
@@ -4483,6 +4506,56 @@ two pointers, ignoring the tags.
...
+
+Returns the GCS pointer of the current thread.
+
+``` c
+  uint64_t __gcspopm(void);
+```
+
+Reads and returns the last entry on the GCS of the current thread and
+updates the GCS pointer to point to the previous entry. If GCS
+protection is disabled then it has no side effect and returns `0`.
```
john-brawn-arm, line 4549: The gcspopm instruction does nothing when gcs is disabled (i.e. the value in the destination register is unchanged). Is the intent that the intrinsic sets the destination register to zero before executing the instruction?

main/acle.md
DanielKristofKiss: I'd prefer to write/read this code in applications:
`if( __chkfeat(_CHKFEAT_GCS)) {
   // do GCS stuff.
}`
chkfeat must clear the bit if the feature is present due to it is a NOP on old CPUs where it won't change the register content. 
That logic twist is not required for the app developers. 


+1 for the 

> Also "feature is available" should be "feature is enabled"

main/acle.md
nsz-arm: yes, i think this is the only way the intrinsic is usable in the "asynchronously disabled gcs" scenario.

(which we don't know yet if linux will support, currently it does not want to)

main/acle.md
```diff
@@ -4506,6 +4506,56 @@ two pointers, ignoring the tags.
...
+enabled then GCS instructions are supported and the code was compiled
+in a GCS protection compatible way (`__ARM_FEATURE_GCS_DEFAULT` was
+defined).
+
+
+## Intrinsics
+
+
+``` c
+  const void *__gcspr(void);
```
nsz-arm, line 4538: i plan to change this back to non-const so that the intrinsics support the gcs write access enabled case (even if that's not expected to be a common setting).

DanielKristofKiss (approved):
LGTM, let see what others think.
NIT maybe the draft could be dropped now from the name of the PR.

nsz-arm:
> NIT maybe the draft could be dropped now from the name of the PR.

potential issues with current spec:
- gcs store intrinsic is missing
- const void* vs void* gcs (modified by bl/ret/... instructions, but normal c language stores don't work on it)
- intrinsic availability check (feature test macros no longer imply intrinsics)

main/acle.md
```diff
@@ -4571,6 +4595,56 @@ two pointers, ignoring the tags.
...
+Each thread has a separate GCS and GCS pointer.
+
+To use the intrinsics, `arm_acle.h` needs to be included.
+
+The intrinsics are only valid to call when GCS instructions are
+supported. The `__chkfeat` intrinsics with `_CHKFEAT_GCS` can be used
+to check if GCS protection is enabled at runtime. If GCS protection is
+enabled then GCS instructions are supported and the code was compiled
+in a GCS protection compatible way (`__ARM_FEATURE_GCS_DEFAULT` was
+defined).
```
john-brawn-arm, line 4620: The final sentence here is a bit weird and I'm not sure what exactly it's saying. I _think_ what it's saying is that "GCS enabled implies GCS available", i.e. if you have ``if (__chkfeat(_CHKFEAT_GCS)) { }`` then you know GCS must be available inside that block, or perhaps the other way around that if you know GCS isn't available then you know ``__chkfeat(_CHKFEAT_GCS)`` will never return true.

john-brawn-arm:
> * gcs store intrinsic is missing
I'm not sure we need this. We can always add it later if there's a need.

> * const void* vs void* gcs (modified by bl/ret/... instructions, but normal c language stores don't work on it)
I think we only really need the pointer to be non-const if we have a gcs store intrinsic. constness is only about if the user is permitted to write to something, so entries on the gcs stack being modified by bl/ret isn't relevant.

> * intrinsic availability check (feature test macros no longer imply intrinsics)
Is this about https://github.com/ARM-software/acle/pull/214? I think the way GCS intrinsics are currently described is OK with regards to that (though I have a couple of comments on the wording).

main/acle.md
nsz-arm: yeah this should be worded better, or removed.

i was trying to point out that the runtime will only enable GCS if all code is built with GCS support, so if __chkfeat check returns true, that implies __ARM_FEATURE_GCS_DEFAULT, no need to check that separately.

john-brawn-arm:
LGTM.

main/acle.md
```diff
@@ -1646,6 +1646,15 @@ mechanisms such as function attributes.
 Pointer Authentication extension (FEAT_PAuth_LR) are available on the target.
 It is undefined otherwise.
 
+### Guarded Control Stack
+
+`__ARM_FEATURE_GCS_DEFAULT` is defined to `1` if the code generation is
+compatible with enabling the Guarded Control Stack (GCS) extension based
+protection.  It is undefined otherwise.
```
DanielKristofKiss, line 1653:
```suggestion
protection. It is undefined otherwise.
```

main/acle.md
```diff
@@ -1646,6 +1646,15 @@ mechanisms such as function attributes.
...
+
+`__ARM_FEATURE_GCS` is defined to `1` if the Guarded Control Stack (GCS)
+extension is available on the target.  It is undefined otherwise.
```
DanielKristofKiss, line 1656:
```suggestion
extension is available on the target. It is undefined otherwise.
```

DanielKristofKiss (approved):
LGTM

yury-khrustalev:
Regarding the return type of `__gcspr()` and `__gcsss()` and the argument type of the latter, it is suggested to implement them as `void *` rather than `const void *` for reasons explained in [this GCC patch's commit message](https://inbox.sourceware.org/gcc-patches/20241108144816.2681175-9-yury.khrustalev@arm.com/).

In a nutshell, the GCS pointer can be writeable in certain situations: via stack switch operation or via the `GCSSTR` instruction.

yury-khrustalev:
This PR is superseded by #364

PR #264, opened 2023-06-13
Fix a number of issues in CMSE examples

### LINKED ISSUE #210: [BUG] CMSE specs: Need examples using new Armv8.1-M instructions to get correct behaviour and show optimal new sequences.
Versions of the CMSE example code that use the new instructions in Armv8.1-M should be created. While some of the instructions (EG CLRM and VSCCLRM) just provide performance and code size improvements, correct usage of the new FPCXT payloads is required to fix an ABI issue. 

For example:
- Many examples (including sections 8.3.1, 8.3.2, 8.3.3, 8.4.2) have a long string of `mov r?, #0` instructions to clear the registers. These can be replaced by the new `CLRM` clear multiple instruction.

- The `FPCXTS` payload should be saved and restored around a Secure -> Non-secure call. This should be incorporated into the example shown in section 8.3.1

- The `FPCXTNS` payload should be saved and restored on entry/exit from a `cmse_nonsecure_entry` function that either uses floating point, or calls another function that may use floating-point. This could be added to a new version of the example in section 8.4.2, although it might be cleaner to have a simpler example that doesn’t also involve stacked arguments.
- Section 8.4.2 contains the following code to detect if an FP context is active, and to clear the registers if they contain secure data:
```
@14: check SFPA bit to see if FP is used
    mrs     r1, control
    tst     r1, #8
    bne     .LdoneFP
    @15: clear floating point caller-saved registers (soft ABI)
    mov     r1, #0
    vmov    s0, s1, r1, r1
    vmov    s2, s3, r1, r1
    ...
    vmov    s30, s31, r1, r1
    @16: clear floating point flags
    vmsr    fpscr, r1
.LdoneFP:
```
This code can be replaced by the new `VSCCLRM` instruction

vhscampos:
- Use CLRM to clear registers instead of a sequence of mov rX, #0.
- Save and restore FPCXTS in transitions from Secure to Non-secure and vice-versa.
- Use VSCCLRM to clear floating-point registers if necessary.

Fix for #210 

---
name: Pull request

---

If this PR is about a bugfix:

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.

### DIFF
--- a/cmse/cmse.md
+++ b/cmse/cmse.md
@@ -160,8 +160,13 @@ Anticipated changes to this document include:
   [Arguments on the stack and floating point handling](#arguments-on-the-stack-and-floating-point-handling).
 * Removed incorrect information about the floating-point ABI used in
   [Arguments on the stack and floating point handling](#arguments-on-the-stack-and-floating-point-handling).
+* Replaced a conditional clearing of floating-point registers by the `VSCCLRM`
+  instruction in the example in
+  [Arguments on the stack and floating point handling](#arguments-on-the-stack-and-floating-point-handling).
 * Corrected description and example in section
   [Non-secure function pointers](#non-secure-function-pointer).
+* Corrected examples to save and restore `FPCXT` in Security state transitions.
+* Replaced sequences of `mov rX, #0` by the `CLRM` instruction in examples.
 
 ## References
 
@@ -1173,7 +1178,8 @@ registers that contain secret information
 > * Clear all registers and flags that have undefined values at the return of a
 >   procedure, according to [[AAPCS]](#AAPCS).
 > * Restore all callee-saved registers as mandated by [[AAPCS]](#AAPCS).
-> * Restore bits [27:0] of FPSCR (Armv8.1-M Mainline only).
+> * Restore the Non-secure Floating-point context payload (`FPCXTNS`)
+> (Armv8.1-M Mainline only).
 
 You can clear the floating-point registers conditionally by checking the 
 `SFPA` bit of the special-purpose `CONTROL` register.
@@ -1266,7 +1272,8 @@ the non-secure state to restore these registers.
 >  * Registers that do not hold secret information.
 > * Clear all registers and flags that have undefined values at the entry to a
 > procedure according to the [[AAPCS]](#AAPCS).
-> * Save and clear bits [27:0] of FPSCR (Armv8.1-M Mainline only).
+> * Save and clear the Secure Floating-point context payload (`FPCXTS`)
+> (Armv8.1-M Mainline only).
 
 A toolchain could provide you with the means to specify that some
 types of variables never hold secret information.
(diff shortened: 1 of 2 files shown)


cmse/cmse.md
```diff
@@ -1808,12 +1810,14 @@ foo:
 __acle_se_foo:
     @ 1: if called from non-secure reserve secure stack space for the arguments
     tst     lr, #1
-    it      eq
+    ittt    eq
     subeq   sp, sp, #8
+    @ 3: save FPCXTNS if called from non-secure. LR is stacked only for alignment purposes
```
696GrocuttT, line 1815: couple of points about this:
1) You shouldn't make this conditional. Just detecting whether we've been called from NS isn't enough to get the correct behaviour as there's some nasty corner cases. To make it easier instructions that access FPCXT_NS already have all the special behaviour built into them, but for that to work they must be executed unconditionally
2) There's no need to use a VMRS followed by a PUSH. Since we expected this to be a common code pattern we created a VSTR (system register) instruction that can access the FPCXT_* payloads. The addressing modes provides by this instruction are flexible enough that you can also preserve the 64bit SP alignment

cmse/cmse.md
```diff
@@ -1842,29 +1846,20 @@ __acle_se_foo:
     bl      bar
     @11: restore used callee-saved registers
     pop     {r4-r6, lr}
-    @12: if called from secure, we are done
+    @12: restore FPCXTNS if called from non-secure
     tst     lr, #1
-    it      ne
+    itte    eq
+    popeq   {r12, lr}
```
696GrocuttT, line 1852: mirror of comments on save of FPCXT above.

cmse/cmse.md
```diff
@@ -1842,29 +1846,20 @@ __acle_se_foo:
...
+    vscclrm {s0-s31}
 .LdoneFP:
-    @17: clear integer caller-saved registers except for return value
-    mov     r1, #0
-    mov     r2, #0
-    mov     r3, #0
+    @15: clear integer caller-saved registers except for return value
+    clrm    {r1-r3}
     @18: clear other registers and the integer status flags
     mov     r12, #0
```
696GrocuttT, line 1864: both of these can also be done by the clrm, just add APSR and R12 to the register list

cmse/cmse.md
```diff
@@ -1723,10 +1728,7 @@ bar:
     @ load the function pointer
     ldr     r4, =foo
     @ clear all integer registers (except for function pointer and arguments)
-    mov     r6,  #0
-    mov     r7,  #0
-    …
-    mov     r12, #0
+    clrm    {r6-r12}
```
696GrocuttT, line 1731: add APSR to register list so you can remove MRS below

cmse/cmse.md
```diff
@@ -1592,20 +1597,23 @@ bar:
...
+    vmrs    r2, fpcxts
+    push    {r2, lr}
     @ save callee-saved integer registers
     push    {r4-r12, lr}
     @ clear all integer registers (except for function pointer and arguments)
-    mov     r2,  #0
-    mov     r3,  #0
-    …
-    mov     r12, #0
+    clrm    {r2-r12}
```
696GrocuttT, line 1606: add APSR to reg list and remove MRS below

cmse/cmse.md
696GrocuttT, line 1: There's a couple of mentions of FPSCR that still need to be updated. EG:
"Save and clear bits [27:0] of FPSCR (Armv8.1-M Mainline only)"

I'm not sure where this has come from, but for v8.1 we should be using the FPCXT payloads to handle FPSCR. For v8.0 manual handling is required, but unless I'm missing something it the whole register must be set to a reasonable value (which may not be zero), and not just bits 27:0

696GrocuttT (changes requested):

cmse/cmse.md
vhscampos: I have replaced occurrences of FPSCR by FPCXT.

rockdreamer:
Closing as this change has been open for a while now with little progress.

PR #276, opened 2023-10-27
Generalise the SME state management attributes

rsandifo-arm:
This patch replaces __arm_shared_za and __arm_new_za with more general attributes.  The purpose is twofold:

* To allow the same approach to be taken for ZT0, and for any other similar state that is added in future.

* To allow the programmer to give the compiler more information about when state is live (if the programmer wants to).

__arm_shared_za is directly equivalent to __arm_inout("za"). Any code that is already using __arm_shared_za, or that prefers that syntax for some reason, could just #define one to the other. Similarly, __arm_new_za is directly equivalent to __arm_new("za").

The patch also removes __arm_preserves_za and replaces it with a more restricted __arm_preserves attribute.  The old  __arm_preserves_za could be attached to both private-ZA and shared-ZA functions. However, the private-ZA version had somewhat dubious semantics:

* It made a promise about how ZA would be handled by a C/C++ function that doesn't use ZA directly.  It wasn't obvious how the burden of keeping that promise was distributed between the programmer and the compiler.  (The feature was always intended to be low-level.)

* The semantics for private-ZA functions meant that callers would still need to prepare a lazy save buffer.  __arm_preserves_za just meant that they could avoid having to restore from it afterwards.

I'm hoping to replace the private-ZA form of __arm_preserves_za with an alternative, optional, extension that avoids the need for the lazy save buffer, and that would be handled entirely by the compiler.

In contrast, __arm_preserves("za") makes a function shared-ZA.

---
name: Pull request

---

If this PR is about a bugfix:

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -348,8 +348,8 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
   [CMSE](#CMSE-ACLE)'s Arguments on the stack and floating point handling.
 * Corrected description and example in [CMSE](#CMSE-ACLE)'s section about
   non-secure function pointers.
-* Added a requirement on [`arm_new_za`](#arm_new_za) to set the initial
-  contents of ZA to zero.
+* Added a requirement on [`arm_new_za`] to set the initial contents of
+  ZA to zero.
 
 #### Changes for next release
 
@@ -369,6 +369,9 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
   for the Armv9.5-A's PAC Enhancements.
 * Introduced a new value to the `__ARM_FEATURE_PAC_DEFAULT` macro to indicate
   the use of PC as a diversifier for [Pointer Authentication](#pointer-authentication).
+* Added a [State management](#state-management) section, replacing the
+  `__arm_shared_za`, `__arm_new_za`, and `__arm_preserves_za` attributes
+  in the previous Alpha SME spec.
 
 ### References
 
@@ -806,9 +809,11 @@ predefine the associated macro to a nonzero value.
 | **Name**                                                    | **Target**            | **Predefined macro**              |
 | ----------------------------------------------------------- | --------------------- | --------------------------------- |
 | [`__arm_locally_streaming`](#arm_locally_streaming)         | function declaration  | `__ARM_FEATURE_LOCALLY_STREAMING` |
-| [`__arm_new_za`](#arm_new_za)                               | function declaration  | `__ARM_FEATURE_SME`               |
-| [`__arm_preserves_za`](#arm_preserves_za)                   | function type         | `__ARM_FEATURE_SME`               |
-| [`__arm_shared_za`](#arm_shared_za)                         | function type         | `__ARM_FEATURE_SME`               |
+| [`__arm_in`](#ways-of-sharing-state)                        | function type         | Argument-dependent                |
+| [`__arm_inout`](#ways-of-sharing-state)                     | function type         | Argument-dependent                |
+| [`__arm_new`](#arm_new)                                     | function declaration  | Argument-dependent                |
+| [`__arm_out`](#ways-of-sharing-state)                       | function type         | Argument-dependent                |
+| [`__arm_preserves`](#ways-of-sharing-state)                 | function type         | Argument-dependent                |
 | [`__arm_streaming`](#arm_streaming)                         | function type         | `__ARM_FEATURE_SME`               |
 | [`__arm_streaming_compatible`](#arm_streaming_compatible)   | function type         | `__ARM_FEATURE_SME`               |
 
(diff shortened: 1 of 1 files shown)


rsandifo-arm:
https://github.com/ARM-software/acle/commit/9d319dbc09f45e9a6877a02d40c02465b4a1a24b updates the patch to include the SME2 intrinsics.  It also tries to clarify that the requirements on inline asms apply to the asm instructions themselves; they're not something that the compiler is supposed to do.

main/acle.md
```diff
@@ -4574,6 +4579,406 @@ two pointers, ignoring the tags.
...
+
+    setup_za();      // ZA is live after this call
+    f_shares_zt0();  // The compiler should save and restore ZA
+                     // around the call ("caller-save")
+    f_shares_za();   // ZA is live before and after this call
+    f_private();     // The compiler should preserve ZA across the call
+                     // It can use the lazy-save mechanism
+    use_za();        // ZA is no longer live after this call
+
+    f_private();     // ZA is not live, no save necessary
```
sdesmalen-arm, line 4978: Could you also describe the liveness of ZT0 (and the need to save/restore it) as part of the example?

main/acle.md
```diff
@@ -4574,6 +4579,406 @@ two pointers, ignoring the tags.
...
+    instructions. This is the default assumption.
+
+*   S is not an input to the asm and S is not an output from the asm.
+    However, the contents of S after executing the asm's instructions might
+    be different from the contents of S before executing the instructions.
+    This is indicated by adding S's string to the asm's clobber list.
+
+    If an asm takes this option for state that is controlled by PSTATE.ZA,
+    the asm itself is responsible for handling the [[AAPCS64]](#AAPCS64)
+    lazy save scheme, as described below.
```
sdesmalen-arm, line 4864: I find this sentence a bit confusing. What exactly is it that is described below? It's not really the lazy-save scheme, as this is referenced but not described.

main/acle.md
```diff
@@ -4574,6 +4579,406 @@ two pointers, ignoring the tags.
...
+
+Otherwise, ZA can be in any state on entry to A if at least one of the
+following is true:
+
+*   F [uses](#uses-state) `"za"`
+
+*   F [uses](#uses-state) `"zt0"`
+
+Otherwise, ZA can be off or dormant on entry to A, as for what AAPCS64
+calls “private-ZA” functions.
```
sdesmalen-arm, line 4889: Can this be phrased the other way around:

* If F uses state S and A's clobber list includes S, then ZA is guaranteed to be active on entry to A.
* If F uses state S and A's clobber list does not include S, then ZA can be in any state on entry to A.
* If F does not use state S, then ZA can be off or dormant on entry to A.

?

main/acle.md
```diff
@@ -8694,117 +9097,32 @@ anchor that can be referred via an internal hyperlink to the paragraph
...
-If a function F1 has [ZA state](#za-state) and it calls a function F2, then:
-
-*   if F2 is a [shared-ZA](#shared-za) function, F1 shares its ZA state
-    with F2.
-
-*   otherwise, F1's ZA state is unchanged by F2.
-
-Again the analogy is with passing or not passing a `pseudo_za_state`
-reference to F2.
+    [`__arm_new("za")`](#arm_new to the function definition.
```
sdesmalen-arm, line 9122:
```suggestion
    [`__arm_new("za")`](#arm_new) to the function definition.
```

main/acle.md
```diff
@@ -4574,6 +4579,406 @@ two pointers, ignoring the tags.
...
+return type for that register.
+
+SVE provides 16 predicate registers, but this number is not exposed to
+C and C++ code. Instead, the program can have as many `svbool_t` objects
+as it needs, and it is the compiler's responsibility to manage the
+register allocation.
+
+However, there are some pieces of architectural state for which this
+approach is not appropriate. For example, [SME's ZA](#za-storage) is a
+single piece of storage: there are not multiple ZAs, and so it does
```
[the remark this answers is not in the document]
rsandifo-arm: I don't think this change is correct.

main/acle.md
rsandifo-arm: The difficulty with breaking it down like this is that the active/dormant/off states depend on both ZA and ZT0 together.  So:

* If F uses state S and A's clobber list does not include S, then ZA can be in any state on entry to A.
* If F does not use state S, then ZA can be off or dormant on entry to A.

is not true for S==ZA if F uses ZT0 and A's clobber list includes ZT0.  In particular, the last point would contradict the first point, since the first point would say ZA must be active (for S==ZT0), whereas the last point would say it must dormant or off (for S==ZA).

So I think it needs to be a chain of if-then-elses that deals with all PSTATE.ZA-based state together.

I can simplify the middle one to:

> Otherwise, ZA can be in any state on entry to A if F [uses](#uses-state) `"za"` or `"zt0"`

if you think that's better.  I wrote it the current way partly for consistency with the first “if”, partly to mimic the rules-based language in the new ISA supplements, and partly to avoid ambiguity about whether “or” includes “and” (which would become harder to deal with if there were ever more than two pieces of state controlled by PSTATE.ZA).

sdesmalen-arm: Ah fair point. In that case I'm happy with the way it's written, I don't think combining those two bullet points into one makes a big difference, and at least it's now consistent with the case above.

main/acle.md
rsandifo-arm: I tried saying various alternative things, but I'm not sure they added much, so in the end I just deleted “, as described below”.

main/acle.md
rsandifo-arm: Good idea.  In the end it seemed simpler to use a separate example, since it became a bit unwieldly otherwise.

sdesmalen-arm (approved):
Thanks for addressing the comments! This PR looks good to me now.

main/acle.md
```diff
@@ -4588,6 +4593,427 @@ two pointers, ignoring the tags.
...
+    f_shares_za();   // ZA is live before and after this call
+    f_private();     // The compiler should preserve ZA across the call
+                     // It can use the lazy-save mechanism
+    use_za();        // ZA is no longer live after this call
+
+    f_private();     // ZA is not live, no save necessary
+  }
+```
+
+ZT0 cannot be lazily saved, so if ZT0 is live before a call to a
```
sallyarmneale, line 4996: saved lazily - don't split the verb

sallyarmneale (commented):
One tiny change

PR #277, opened 2023-11-21
[FMV] Change the default versions name mangling.

DanielKristofKiss:
The default version's name to be mangled to avoid problems with ifuncs due to the ifunc to be called as the base name of the function.

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -2490,8 +2490,9 @@ compiler and it is enabled.
 
 ### Name mangling
 
-The `"default"` version is not mangled on top of the language-specific name
-mangling.
+The `"default"` version is mangled with `".default"` on top of the
+language-specific name mangling. When the `"default"` version matches with an
+explicitly provided version an alias to be generated.
 
 The mangling function is compatible with the mangling for version information of
 the [[cxxabi]](#cxxabi), and it is defined as follows:
@@ -2510,7 +2511,7 @@ For example:
 __attribute__((target_clones("crc32", "aes+sha1")))
 int foo(){..}
 ```
-will produce these mangled names for C language: `foo`, `foo._Mcrc32`,
+will produce these mangled names for C language: `foo.default`, `foo._Mcrc32`,
 `foo._Msha1Maes`.
 
 
--- a/main/acle.md
+++ b/main/acle.md
@@ -2493,8 +2493,8 @@ compiler and it is enabled.
 
 The `"default"` version is mangled with `".default"` on top of the
 language-specific name mangling.
-language-specific name mangling. When the `"default"` version matches with an
-explicitly provided version an alias to be generated.
+All versioned functions with their mangled name are always resolvable.
+A function is expected to be resolvable with the original mangled name of the function.
 
 The mangling function is compatible with the mangling for version information of
 the [[cxxabi]](#cxxabi), and it is defined as follows:
@@ -2514,7 +2514,8 @@ __attribute__((target_clones("crc32", "aes+sha1")))
 int foo(){..}
 ```
 will produce these mangled names for C language: `foo.default`, `foo._Mcrc32`,
-`foo._Msha1Maes`.
+`foo._Msha1Maes` while `foo` is a callable external symbol which leads to one
+of the versioned function.
 
 
 ### Mapping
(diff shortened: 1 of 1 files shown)


vhscampos:
Some points aren't clear in the text:
1. Will this new `default` name mangling happen only with FMV enabled or globally?
2. If the _only_ version specified is the `default` one, either explicitly with the attribute or implictly without the attribute, will the new mangling still take place?

DanielKristofKiss:
> Some points aren't clear in the text:
> 
> 1. Will this new `default` name mangling happen only with FMV enabled or globally?

Only when FMV enabled. By amending the `default` version with `"default"` the basename could be used by the resolved symbol, referenced externally as it won't collide with the default function.

> 2. If the _only_ version specified is the `default` one, either explicitly with the attribute or implictly without the attribute, will the new mangling still take place?

Depends on, for the same as above reasons. if the toolchain can avoid emitting redirections then no need to use the default version. maybe a alias could be useful.

main/acle.md
```diff
@@ -2474,8 +2476,9 @@ compiler and it is enabled.
 
 ### Name mangling
 
-The `"default"` version is not mangled on top of the language-specific name
-mangling.
+The `"default"` version is mangled with `".default"` on top of the
+language-specific name mangling. When the `"default"` version matches with an
+explicitly provided version an alias to be generated.
```
DanielKristofKiss, line 2481:
```suggestion
explicitly provided version an alias to be generated.
All versions expected to have at least alias that points to actual version.
```

andrewcarlotti:
I don't understand the remark about aliases in this change. Is this saying that the compiler can optimise versions with identical function bodies to a single function with additional aliased names? If so, I don't think that's relevant to this spec. Alternatively, is it referring to a case where a non-default version specifies only features that the compiler is already assuming to be present?

andrewcarlotti:
Aside from that, I think the mangling changes are reasonable. It might be worth adding an explicit remark that the versioned symbol (i.e. the one that is resolved to one of the versions) should have the same mangling as would be used if the function didn't use function multiversioning.

DanielKristofKiss:
> I don't understand the remark about aliases in this change. Is this saying that the compiler can optimise versions with identical function bodies to a single function with additional aliased names? If so, I don't think that's relevant to this spec. Alternatively, is it referring to a case where a non-default version specifies only features that the compiler is already assuming to be present?

The intention is to be sure all function names remain always available when the certain optimisation eliminates a version. (e.g. the default and a given version got merged).

> Aside from that, I think the mangling changes are reasonable. It might be worth adding an explicit remark that the versioned symbol (i.e. the one that is resolved to one of the versions) should have the same mangling as would be used if the function didn't use function multiversioning.

ACK.

main/acle.md
```diff
@@ -2496,8 +2499,9 @@ For example:
 __attribute__((target_clones("crc32", "aes+sha1")))
 int foo(){..}
 ```
-will produce these mangled names for C language: `foo`, `foo._Mcrc32`,
-`foo._Msha1Maes`.
+will produce these mangled names for C language: `foo.default`, `foo._Mcrc32`,
+`foo._Msha1Maes` while `foo` is a callable external symbol which leads to one
+of the versioned function.
```
vhscampos, line 2504: Nit: "versioned function" -> "version functions". Plural

main/acle.md
```diff
@@ -2476,8 +2477,10 @@ compiler and it is enabled.
 
 ### Name mangling
 
-The `"default"` version is not mangled on top of the language-specific name
-mangling.
+The `"default"` version is mangled with `".default"` on top of the
+language-specific name mangling.
+All versioned functions with their mangled name are always resolvable.
```
vhscampos, line 2482: Nit: "name" -> "names". Plural

vhscampos, line 2482: I suggest that you move this line to the end of the previous one, i.e. make the two form a single paragraph.

vhscampos (changes requested):

andrewcarlotti:
This looks good to me now. I've already implemented this mangling in GCC.

vhscampos (approved):

sallyarmneale (commented):
Minor changes

PR #301, opened 2024-02-23
[FMV] Add __FUNCTION_MULTI_VERSIONING_SUPPORT_LEVEL.

DanielKristofKiss:

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -400,8 +400,8 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * Added a requirement for function version declaration in Function Multi Versioning.
 * Fixed some rendering issues in the online Markdown documentation and fixed
   a misplaced anchor.
-* Added __FUNCTION_MULTI_VERSIONING_SUPPORT_LEVEL to indicate the support level of the
-  [Function Multi Versioning](#function-multi-versioning).
+* Added `__FUNCTION_MULTI_VERSIONING_SUPPORT_LEVEL` to indicate the support
+  level of the [Function Multi Versioning](#function-multi-versioning).
 
 ### References
 
@@ -2540,7 +2540,7 @@ compiler and it is enabled.
 `__FUNCTION_MULTI_VERSIONING_SUPPORT_LEVEL` is defined to the currently supported
 version of the ACLE. The value and the format are the same as the `__ARM_ACLE`.
 
-For example could be implemted as:
+For example, it can be implemented as:
 
 ``` c
 #define __FUNCTION_MULTI_VERSIONING_SUPPORT_LEVEL __ARM_ACLE_VERSION(2024, 3, 0)
--- a/main/acle.md
+++ b/main/acle.md
@@ -400,6 +400,8 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * Added a requirement for function version declaration in Function Multi Versioning.
 * Fixed some rendering issues in the online Markdown documentation and fixed
   a misplaced anchor.
+* Added __FUNCTION_MULTI_VERSIONING_SUPPORT_LEVEL to indicate the support level of the
+  [Function Multi Versioning](#function-multi-versioning).
 
 ### References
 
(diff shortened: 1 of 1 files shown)


DanielKristofKiss:
cc: @ilinpv @jroelofs @andrewcarlotti

ilinpv:
Looks reasonable to me.

jroelofs:
Ditto. Thank you!

DanielKristofKiss:
Hi @vhscampos,
One of the reasons:
https://github.com/llvm/llvm-project/issues/79659 

If the macro reports the version then a program code could check the support level. Later if we introduce new features to the spec then we would have a way to detect it.
Also make easier to check feature level support to ( .e.g. arch FeatureX only introduced in 2025 release of ACLE )

vhscampos:
I'm not sure this is a good idea. `__HAVE_FUNCTION_MULTI_VERSIONING` is a name that reflects a binary choice. Even having it as a bitmask would be a stretch (the previously suggested approach in https://github.com/llvm/llvm-project/issues/79659).

If we want to gate things behing conditions such as this:
`#if __HAVE_FUNCTION_MULTI_VERSIONING >= __ARM_ACLE_VERSION(2024, 1, 0 )`
Then we should just use directly:
`#if __ARM_ACLE >= __ARM_ACLE_VERSION(2024, 1, 0 )`

jroelofs:
I am not sure that using `__ARM_ACLE >= __ARM_ACLE_VERSION(2024, 1, 0 )` directly solves the problem that I have:

> the presence of compilers in the wild that claim to but don't actually support [FMV] means we don't have a great way to gate usage of this feature via the preprocessor.

For context, in clang, support for FMV depends on:
* platform support for IFuncs, or an equivalent lowering (which isn't implemented for all platforms)
* libcompiler_rt being the runtime library

which are both orthogonal to the implemented ACLE version.

vhscampos:
Sorry for the long delay. I had some discussions internally but the outcome got lost in the process.

I reckon that __HAVE_FUNCTION_MULTI_VERSIONING is a macro whose value should remain a boolean. Its name has the explicit meaning of a boolean piece of information.

For the purpose intended in this change, I recommend that a new macro is created with a name similar to __MULTI_VERSIONING_SUPPORT_LEVEL, with a value derived from the relevant ACLE version:

`#define __MULTI_VERSIONING_SUPPORT_EVEL __ARM_ACLE_VERSION(2024, 1, 0)`

jroelofs:
works for me

vhscampos:
@labrinea Can you please review this as well once you're able to?

labrinea:
Lgtm

main/acle.md
```diff
@@ -2535,6 +2537,15 @@ following:
 versioning mechanism described in this section is supported by the
 compiler and it is enabled.
 
+`__FUNCTION_MULTI_VERSIONING_SUPPORT_LEVEL` is defined to the currently supported
+version of the ACLE. The value and the format are the same as the `__ARM_ACLE`.
+
+For example could be implemted as:
```
vhscampos, line 2543: Typo "implemted"

main/acle.md
```diff
@@ -400,6 +400,8 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * Added a requirement for function version declaration in Function Multi Versioning.
 * Fixed some rendering issues in the online Markdown documentation and fixed
   a misplaced anchor.
+* Added __FUNCTION_MULTI_VERSIONING_SUPPORT_LEVEL to indicate the support level of the
```
vhscampos, line 403: Please wrap the macro name between ticks for better formatting

main/acle.md
vhscampos, line 2543: Better phrasing: "For example, it can be implemented as:"

vhscampos (changes requested):

main/acle.md
DanielKristofKiss: done

main/acle.md
DanielKristofKiss: done

DanielKristofKiss: done

vhscampos (approved):
LGTM. Thank you

sallyarmneale (commented):
LGTM

PR #315, opened 2024-03-28
Remove useless function multiversioning features

alicecarlotti:
Supporting a feature in function multiversioning requires that the rest of the toolchain can support enabling the feature.  In GCC/Binutils, this requires either that there is an equivalent feature extension available in the `-march` command line option, or that the feature introduces no new instructions (and therefore does not need to be enabled when passing to the assembler).  However, many of the features listed in in the original function multiversioning specification do not meet this criteria. These fall into four categories:

1. Features that were originally linked to a specific architecture version (fcma, jscvt, frintts, flagm2, wfxt, rcpc2): these should get their own flags across the toolchain (and I've already added this support in Binutils).

2. Features that are combined with other features in existing command line options: these should also be merged in the function multiversioning specification.

3. Features that indicate support for hint instructions: these can be dropped, since the instructions can be used unconditionally.

4. Features that enable existing instruction behaviour to be changed when a system register flag is set: the function multiversioning resolvers don't check for runtime enablement of the control flags, so this isn't a suitable way of exploiting the behaviour enabled by those flags.

5. Features that can also be expressed as a combination of two other features.

We therefore remove support for the following features from the specification:

- sha1 (2): included within +sha2
- pmull (2): included within +aes
- dit (4)
- dgh (3)
- ebf16 (4)
- sve-bf16 (5)
- sve-ebf16 (4)
- sve-i8mm (5)
- sve2-pmull128 (2): included within +sve2-aes
- memtag2 (2): included within +memtag
- memtag3 (4)
- ssbs2 (2): included within +ssbs
- bti (3)
- ls64_v, ls64_accdata (2): included within +ls64

---
name: Pull request

---

If this PR is about a bugfix:

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -375,7 +375,14 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
   * Change name mangling of the default version.
   * Align priorities to account for feature dependencies.
   * Introduce alternative names (aliases) `rdma` for `rdm`.
-  * Correct FEAT_BTI feature register value.
+  * Combine some features, to align more closely with compiler command line
+    options (eliminating sha1, pmull, sve2-pmull128, memtag2, ssbs2, ls64_v
+    and ls64_accdata).
+  * Remove features that are combinations of other features (sve-bf16 and
+    sve-i8mm).
+  * Remove features that only enable hint instructions (dgh and bti).
+  * Remove features that indicate only that a specific behaviour can be
+    separately switched on (dit, ebf16, sve-ebf16 and memtag3).
 * Introduced the `__ARM_FEATURE_PAUTH_LR` feature macro in section
   [Pointer Authentication](#pointer-authentication) to indicate target support
   for the Armv9.5-A's PAC Enhancements.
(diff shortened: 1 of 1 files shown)


labrinea:
All the suggested changes LGTM. Another feature in question that I believe falls in the second category is "ssbs" (included within +ssbs2) and one that falls in the forth category is "rpres".

labrinea:
> All the suggested changes LGTM. Another feature in question that I believe falls in the second category is "ssbs" (included within +ssbs2) and one that falls in the forth category is "rpres".

Oh I realised ssbs was already on your list, I must have missed it.

main/acle.md
```diff
@@ -2606,35 +2610,23 @@ The following table lists the architectures feature mapping for AArch64
...
    | 430           | `FEAT_SME`               | sme           | ```ID_AA64PFR1_EL1.SME == 0b0001```       |
-   | 440           | `FEAT_MTE`               | memtag        | ```ID_AA64PFR1_EL1.MTE >= 0b0001```       |
-   | 450           | `FEAT_MTE2`              | memtag2       | ```ID_AA64PFR1_EL1.MTE >= 0b0010```       |
-   | 460           | `FEAT_MTE3`              | memtag3       | ```ID_AA64PFR1_EL1.MTE >= 0b0011```       |
+   | 440           | `FEAT_MTE`,`FEAT_MTE2`   | memtag        | ```ID_AA64PFR1_EL1.MTE >= 0b0010```       |
    | 470           | `FEAT_SB`                | sb            | ```ID_AA64ISAR1_EL1.SB == 0b0001```       |
    | 480           | `FEAT_SPECRES`           | predres       | ```ID_AA64ISAR1_EL1.SPECRES == 0b0001```  |
-   | 490           | `FEAT_SSBS`              | ssbs          | ```ID_AA64PFR1_EL1.SSBS == 0b0001```      |
-   | 500           | `FEAT_SSBS2`             | ssbs2         | ```ID_AA64PFR1_EL1.SSBS == 0b0010```      |
-   | 510           | `FEAT_BTI`               | bti           | ```ID_AA64PFR1_EL1.BT == 0b0001```        |
```
DanielKristofKiss, line 2634: Please don't drop BTI it is used to detect the feature and generate different code/functionality based on it. 
see: https://chromium-review.googlesource.com/c/chromium/src/+/5453915

DanielKristofKiss (changes requested):

DanielKristofKiss:
"Features that indicate support for hint instructions: these can be dropped, since the instructions can be used unconditionally." 
but still the functionality may matters. FMV could be used for other stuff beside codegen.

andrewcarlotti:
> All the suggested changes LGTM. Another feature in question that I believe falls in the second category is "ssbs" (included within +ssbs2) and one that falls in the forth category is "rpres".

Ah - I hadn't noticed that FEAT_RPRES only applied when FEAT_AFP Alternate Handling is enabled. That probably put it in my fourth category as well, although it's slightly less clear cut than the other cases.

main/acle.md
andrewcarlotti: So, as I understand it, that use chromium example is using FMV on a function that converts a PageAccessibilityConfiguration into an integer mask that will be passed to mmap calls. The FMV usage is to determine whether the PROT_BTI and PROT_MTE bits should be set in the mask. The main motivation for switching to FMV in this patch is that caching the feature identification check in the GOT is supposedly more resistant to modification by an attacker global structure.

One thing that's not clear to me is whether the check is necessary at all. It looks like the Linux kernel might ignore just PROT_MTE when the system doesn't support it, so maybe it's safe to always pass PROT_MTE in the bitmask. Perhaps the same is true for PROT_BTI. However, I wouldn't fully trust my limited code reading, so might be mistaken.

It's possible to support BTI without adding new command line options, so I'm not particularly opposed to retaining this feature if it is actually useful. It originally felt like there ought to be a better way to handle this use case, but on further consideration approach seems reasonable (assuming that the kernel requires the PROT_BTI and PROT_MTE bits to be unset).

andrewcarlotti: Ah - it does check:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/arm64/include/asm/mman.h#n38
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/mm/mprotect.c#n706

Wilco1:
Can we progress this? It seems there is agreement on most of these, so just keep FEAT_BTI.

labrinea:
Ping @DanielKristofKiss @rsandifo-arm

DanielKristofKiss:
@labrinea I think we need new version of this PR to reflect above comments and also has merge conflicts.

vhscampos:
Hi all, can you please confirm if this patch has already been approved? If it's just a matter of resolving merge conflicts, I can do it

labrinea:
Hey Victor. No, it hasn't been approved. It is still work in progress. Cheers.

jroelofs:
> 4. Features that enable existing instruction behaviour to be changed when a system register flag is set: the function multiversioning resolvers don't check for runtime enablement of the control flags, so this isn't a suitable way of exploiting the behaviour enabled by those flags.

In the case of `bti` this conflates the status of the flag with the test for the presence of the ISA feature. You're right that FMV isn't appropriate for a status check, but it _is_ useful for the ISA feature presence check, which for `dit` you need around your `msr dit, x0`'s. I'm not sure about `memtag4` and/or `ebf16`.

labrinea:
This ticket is outdated. The issue has been addressed by separate PRs and now can be closed. The current ACLE is up to date, you may rely on that.

rockdreamer:
Closing as outdated.

PR #319, opened 2024-04-24
Add new requirements to code generation of CMSE Entry function and Non-secure calls

pratlucas:
This addresses the vulnerability described in CVE-2024-0151.

Patch by Victor Campos.

Co-authored-by: Victor Campos <victor.campos@arm.com>

### DIFF
--- a/cmse/cmse.md
+++ b/cmse/cmse.md
@@ -1155,14 +1155,17 @@ declared type's range in an attempt to cause out-of-bounds memory accesses.
 
 <span id="requirement-47" class="requirement-box"></span>
 > A compiler generating code for an entry function must, for each parameter
-> that is an integral Fundamental Data Type smaller than a word, narrow the
-> parameter to a value within its declared type's range.
+> that is an integral Fundamental Data Type smaller than a word, make no
+> assumptions about the value of the padding bits, even when the value of
+> those bits are defined by the AAPCS.
 
 A possible implementation is shown below:
 
 ```c
+int array[256];
+
 __attribute__((cmse_nonsecure_entry))
-int func(int *array, unsigned short idx) {
+int func(unsigned char idx) {
     return array[idx];
 }
 ```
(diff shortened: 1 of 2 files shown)


cmse/cmse.md
```diff
@@ -1138,9 +1145,45 @@ caller stack frame](#figure5).
...
+Fundamental Data Types smaller than a word to 4 bytes. An Entry function must
+not assume that callers follow this rule, that is, it cannot presume that
+integral parameters will have been zero- or sign-extended to 4 bytes. For
+example, an attacker might create code that passes arguments out of their
+declared type's range in an attempt to cause out-of-bounds memory accesses.
 
 <span id="requirement-47" class="requirement-box"></span>
+> A compiler generating code for an entry function must, for each parameter
+> that is an integral Fundamental Data Type smaller than a word, narrow the
+> parameter to a value within its declared type's range.
```
momchil-velikov, line 1159: Are there other ways to "narrow" the parameter other than sign- or zero- extending it? If not, I would suggest rephrasing in terms of sign-extending/zero-extending to 4-bytes, as it looks more clear.

cmse/cmse.md
```diff
@@ -1138,9 +1145,45 @@ caller stack frame](#figure5).
...
+func:
+@ narrow 'idx' to 16 bits before first use
+uxth    r1, r1
+ldr.w   r0, [r0, r1, lsl #2]
+...
+bxns lr
+```
+
+We recommend that function parameters with integral types smaller than 4
+bytes should be avoided. This guidance extends to underlying types of `enum`
```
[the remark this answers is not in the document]
smithp35: I think enums are mentioned "This guidance extends to underlying types of `enum`" as `-fshort-enums` are affected.
structs and unions are not affected (not integral types). The AAPCS does not require these to be sign-extended by the caller when passed as parameters or returned by callees.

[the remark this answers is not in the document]
ostannard: It's not clear to me whether the second sentence of this paragraph is (1) a requirement on the compiler, or (2) a continuation of the non-normative advice to programmers in the first sentence. The word "guidance" suggests it's (2), but this probably was intended to  be (1), so moving it elsewhere would clarify that.

cmse/cmse.md
vhscampos: In discussions held previously, it was mentioned that saying truncate+extend explicitly could be overprescriptive. A compiler could for example decide for a check-and-abort implementation instead, which would be as valid.

vhscampos: But in line with one of @ostannard's comments, one can argue that the current form may already be overprescriptive.

cmse/cmse.md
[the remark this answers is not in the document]
vhscampos: It's indeed a continuation of (2). I'll replace "guidance" by "recommendation" to make it clearer.

cmse/cmse.md
```diff
@@ -1138,9 +1145,51 @@ caller stack frame](#figure5).
...
+@ narrow 'idx' to 8 bits before first use
+uxtb	r0, r0
+movw	r1, :lower16:array
+movt	r1, :upper16:array
+ldr.w	r0, [r1, r0, lsl #2]
+...
+bxns lr
+```
+
+We recommend that function parameters with integral types smaller than 4
```
ostannard, line 1186: What is the reason for this recommendation? Are there other potential problems with small types, or is this because there are existing compilers which don't follow these rules?

ostannard (approved):
LGTM

sallyarmneale (commented):
Looks good to me.

PR #321, opened 2024-05-29
Remove request for preprocessor guards from header files.

### LINKED ISSUE #340: [BUG] Follow up on #321
Pull Request #321 was merged with a few nits to be addressed. Also, it did not include a changelog entry.

This is a blocker for ACLE release 2024Q3.

CarolineConcatto:
---
name: Pull request

---

If this PR is about a bugfix:

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -1074,24 +1074,23 @@ header, but this might limit function multi-versioning capabilities:
 
 Including `<arm_sme.h>` also includes [`<arm_sve.h>`](#arm_sve.h).
 
-These macro features:
-`__ARM_ACLE`, `__ARM_FEATURE_FP16_SCALAR_ARITHMETIC`, `__ARM_FEATURE_BF16`,
-`__ARM_NEON`, `__ARM_FEATURE_SVE` and ` __ARM_FEATURE_SME`
-represent a compiler's ability to use the instructions associated
-with the feature via ACLE documented builtins and within inline assembly.
-They may also be influenced by source code mechanisms like `#pragma` or
-`attribute`, but users should not rely on this. They are defined for user
-convenience.
+### Predefined Macros and Headers
+
+The macros can represent a compiler's ability to use the instructions
+associated with the feature via ACLE documented intrinsics and within inline
+assembly. However users should not rely solely on macro definition, because
+there are intrinsics that may be influenced by mechanisms like `#pragma` or
+`attribute`. The macros are defined only for user convenience.
 
 For example:
 
 ``` c
    #ifdef __ARM_FEATURE_SME
    #include<arm_sme.h>
+     void foo(svbool_t pg, void *ptr, uint32_t slice_base) {
+        svst1_hor_za8(0, slice_base, pg, ptr);
+      }
    #endif
-   void foo(svbool_t pg, void *ptr, uint32_t slice_base) {
-      svst1_hor_za8(0, slice_base, pg, ptr);
-   }
 ```
 
 `foo`  uses `__ARM_FEATURE_SME` to have available SME intriniscs,
--- a/main/acle.md
+++ b/main/acle.md
@@ -1085,10 +1085,14 @@ Including `<arm_sme.h>` also includes [`<arm_sve.h>`](#arm_sve.h).
 
 ### Predefined feature macros and header files
 
-Testing a feature macro returns the compiler's availability for the feature via ACLE
-documented intrinsics and inline assembly. This property can be independent of
-where the test is performed, meaning a feature macro does not guarantee the
-validity of using the feature at the point it is tested. For example:
+Feature macros indicate whether that feature's intrinsics and inline assembly instructions
+are available in normal unannotated functions. In some implementations, the macro definitions
+are invariant throughout a translation unit, while in other implementations, the macro
+definitions can vary with pragmas.
+
+The macro definitions are otherwise independent of where the test is performed,
+meaning that a feature macro does not guarantee that the feature can be used at the
+point that it is tested. For example:
 
 ``` c
     #include <arm_sme.h>
@@ -1100,10 +1104,9 @@ validity of using the feature at the point it is tested. For example:
     }
 ```
 
-Whilst testing `__ARM_FEATURE_SME` ensures the compiler has available the SME
-intrinsic `svst1_hor_za8`, `foo` will fail to compile because it has not been
-decorated with the necessary keywords to guarantee the function will be
-executed in streaming mode.
+While testing `__ARM_FEATURE_SME` ensures that the SME intrinsic `svst1_hor_za8`
+is available, `foo` will fail to compile because the call does not occur in a
+[streaming statement](#streaming-statement).
 
 Not all such issues can be caught during compilation and may instead result in
 runtime failures.
--- a/main/acle.md
+++ b/main/acle.md
@@ -1075,10 +1075,9 @@ Including `<arm_sme.h>` also includes [`<arm_sve.h>`](#arm_sve.h).
 
 ### Predefined feature macros and header files
 
-Testing a feature macro returns the compiler's availability for the feature
-via ACLE documented intrinsics and inline assembly. This property can be
-independent of where the test is performed. There should be no assumption on
-the order or context in wich the preprocessor macros are evaluated.
+Evaluating a feature macro returns the availability of intrinsics and inline
+assembly for that feature, but no assumptions should be made on the order or
+context in which the preprocessor macros are evaluated.
 
 The compiler may add additional restrictions to the intrinsics beyond what is
 captured by the ACLE macros depending on the context in which the intrinsics
@@ -1093,8 +1092,8 @@ are used. For example:
     }
 ```
 
-While testing `__ARM_FEATURE_SME` ensures that the SME intrinsic `svst1_hor_za8`
-is available, `foo` will fail to compile because the call does not occur in a
+If `__ARM_FEATURE_SME` evaluates to `true` the SME intrinsic `svst1_hor_za8`
+is available, but `foo` may still fail to compile because the call does not occur in a
 [streaming statement](#streaming-statement).
 
 ## Attributes
--- a/main/acle.md
+++ b/main/acle.md
@@ -1003,7 +1003,8 @@ Including `<arm_sve.h>` also includes the following header files:
 
 `<arm_neon_sve_bridge.h>` defines intrinsics for moving data between
 Neon and SVE vector types; see [NEON-SVE Bridge](#neon-sve-bridge)
-for details.
+for details. The `__ARM_NEON_SVE_BRIDGE` macro should be tested
+-before including the header:
 
 ``` c
   #ifdef __ARM_NEON_SVE_BRIDGE
--- a/main/acle.md
+++ b/main/acle.md
@@ -925,8 +925,9 @@ and:
 to the more specific header files below. These intrinsics are in the
 C implementation namespace and begin with double underscores. It is
 unspecified whether they are available without the header being
-included. The `__ARM_ACLE` macro can be tested before including the
-header, but this might limit function multi-versioning capabilities:
+included. The `__ARM_ACLE` macro can be tested before including the header.
+However, this is not recommended for cases where target feature selection is
+overridden within code, for example, when using #function-multi-versioning.
 
 ``` c
   #ifdef __ARM_ACLE
@@ -939,9 +940,10 @@ header, but this might limit function multi-versioning capabilities:
 `<arm_fp16.h>` is provided to define the scalar 16-bit floating point
 arithmetic intrinsics. As these intrinsics are in the user namespace,
 an implementation would not normally define them until the header is
-included. The `__ARM_FEATURE_FP16_SCALAR_ARITHMETIC` macro can be
-tested before including the header, but this might limit function
-multi-versioning capabilities:
+included. The `__ARM_FEATURE_FP16_SCALAR_ARITHMETIC` macro can be tested before
+including the header. However, this is not recommended for cases where target
+feature selection is overridden within code, for example, when using
+#function-multi-versioning.
 
 ``` c
   #ifdef __ARM_FEATURE_FP16_SCALAR_ARITHMETIC
@@ -954,8 +956,10 @@ multi-versioning capabilities:
 `<arm_bf16.h>` is provided to define the 16-bit brain floating point
 arithmetic intrinsics. As these intrinsics are in the user namespace,
 an implementation would not normally define them until the header is
-included. The `__ARM_FEATURE_BF16` macro can be tested before including
-the header, but this might limit function multi-versioning capabilities:
+included. The `__ARM_FEATURE_BF16` macro can be tested before including the
+header. However, this is not recommended for cases where target feature
+selection is overridden within code, for example, when using
+function-multi-versioning.
 
 ``` c
   #ifdef __ARM_FEATURE_BF16
@@ -977,8 +981,10 @@ intrinsics](#advanced-simd-neon-intrinsics) and associated
 [data types](#vector-data-types). As these intrinsics and data types are
 in the user namespace, an implementation would not normally define them
 until the header is included. The `__ARM_NEON` macro can be tested before
-including the header, but this might limit function multi-versioning
-capabilities:
+including the header. However, this is not recommended for cases where target
+feature selection is overridden within code, for example, when using
+#function-multi-versioning.
+
 
 ``` c
   #ifdef __ARM_NEON
@@ -1001,8 +1007,9 @@ following it. --><span id="arm_sve.h"></span>
 `<arm_sve.h>` defines data types and intrinsics for SVE and its
 extensions; see [SVE language extensions and
 intrinsics](#sve-language-extensions-and-intrinsics) for details.
-The `__ARM_FEATURE_SVE` macro can be tested before including the header,
-but this might limit function multi-versioning capabilities:
+The `__ARM_FEATURE_SVE` macro can be tested before including the header.
+However, this is not recommended for cases where target feature selection is
+overridden within code, for example, when using #function-multi-versioning.
 
 ``` c
   #ifdef __ARM_FEATURE_SVE
@@ -1064,7 +1071,9 @@ change or be extended in the future.
 `<arm_sme.h>` declares functions and defines intrinsics for SME
 and its extensions; see [SME language extensions and intrinsics](#sme-language-extensions-and-intrinsics)
 for details. The `__ARM_FEATURE_SME` macro can be tested before including the
-header, but this might limit function multi-versioning capabilities:
+header. However, this is not recommended for cases where target feature
+selection is overridden within code, for example, when using
+#function-multi-versioning.
 
 ``` c
   #ifdef __ARM_FEATURE_SME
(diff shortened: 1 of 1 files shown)


main/acle.md
```diff
@@ -1019,13 +1003,10 @@ Including `<arm_sve.h>` also includes the following header files:
 
 `<arm_neon_sve_bridge.h>` defines intrinsics for moving data between
 Neon and SVE vector types; see [NEON-SVE Bridge](#neon-sve-bridge)
-for details.  The `__ARM_NEON_SVE_BRIDGE` macro should be tested
-before including the header:
+for details.
 
 ``` c
-  #ifdef __ARM_NEON_SVE_BRIDGE
```
paulwalker-arm, line 1026: I believe this should remain because it's a test if the compiler supports this software extension rather than something related to a target feature.

main/acle.md
```diff
@@ -1072,6 +1074,33 @@ including the header:
 
 Including `<arm_sme.h>` also includes [`<arm_sve.h>`](#arm_sve.h).
 
+These macro features:
+`__ARM_ACLE`, `__ARM_FEATURE_FP16_SCALAR_ARITHMETIC`, `__ARM_FEATURE_BF16`,
+`__ARM_NEON`, `__ARM_FEATURE_SVE` and ` __ARM_FEATURE_SME`
+represent a compiler's ability to use the instructions associated
+with the feature via ACLE documented builtins and within inline assembly.
+They may also be influenced by source code mechanisms like `#pragma` or
+`attribute`, but users should not rely on this. They are defined for user
```
mgabka, line 1083: this is badly phrased.  Do you mean here that users should not rely on #pragma or attribute or the feature.
macros?

If it is about the macros, I think we need to phrase it differently and explain that the presence of the macro does not guarantee that the architecture feature is always present at various stages of compilation or execution.

main/acle.md
```diff
@@ -1072,6 +1074,33 @@ including the header:
...
+convenience.
+
+For example:
+
+``` c
+   #ifdef __ARM_FEATURE_SME
+   #include<arm_sme.h>
+   #endif
+   void foo(svbool_t pg, void *ptr, uint32_t slice_base) {
+      svst1_hor_za8(0, slice_base, pg, ptr);
```
mgabka, line 1093: don't you need to extend the endif to cover the definition of foo?

main/acle.md
CarolineConcatto: Hi Maciej,
I re-wrote this paragraph and added a section at the end  of HEADER to explain headers + macros. 
Just in case the ACLE has a section: **Feature test macros**. It is  generic, it is about all predefined macros. 
The point here is that  the ACLE/I  want to explain that  headers protected by macro  does not imply we can always use intrinsics.

main/acle.md
```diff
@@ -1072,6 +1074,32 @@ including the header:
...
+   #endif
+```
+
+`foo`  uses `__ARM_FEATURE_SME` to have available SME intriniscs,
+like `svst1_hor_za8`. However it has the compilation error:
+`builtin can only be called from a streaming function`
+for the command line:
+`armclang -O3 -target aarch64-linux-gnu -march=armv8-a+sme`
+that defines `__ARM_FEATURE_SME` to one.
+
```
paulwalker-arm, line 1102: How about?

Testing a feature macro returns the compiler's support for the feature via ACLE documented intrinsics and inline assembly. This property can be independent of where the test is performed, meaning a feature macro does not guarantee the validity of using the feature at the point it is tested. For example:

``` c
    #include <arm_sme.h>

    void foo(svbool_t pg, void *ptr, uint32_t slice_base) {
    #ifdef __ARM_FEATURE_SME
      svst1_hor_za8(0, slice_base, pg, ptr);
    #endif
    }
```

Whilst testing \`__ARM_FEATURE_SME\` ensures the compiler supports the SME intrinsic \`svst1_hor_za8\`, \`foo\` will fail to compile because it has not been decorated with the necessary keywords to guarantee the function will be executed in streaming mode.

Not all such issues can be caught during compilation and may instead result in runtime failures.

main/acle.md
```diff
@@ -1072,6 +1074,32 @@ including the header:
 
 Including `<arm_sme.h>` also includes [`<arm_sve.h>`](#arm_sve.h).
 
+### Predefined Macros and Headers
```
mgabka, line 1077: should this be changed to : "### predefined feature macros and header files" ?

main/acle.md
mgabka: > Testing a feature macro returns the compiler's support for the feature via ACLE documented intrinsics and inline assembly.

Not sure if it is just me, the key is how we define "support".
For me compiler supports ACLE when the given ACLE is implemented. While the feature macro is only defined when appropriate features are enabled in the cmd line option only. Maybe worth to change it to sth like:

"Testing a feature macro allows to check if the compiler can make use of the architecture feature via ACLE documented intrinsics and inline assembly assuming the provided compilation flags."

What do you think?

paulwalker-arm: This distinction is not relevant because there is no assumption about compilation flags.  The macro alone is sufficient to detect at the source level whether the compiler (in the mode it's in when compiling the input) can support a specific feature.  By support we effectively mean "is able to compile the input when it is correctly formed".

The provided example shows something that is not "correctly formed" where the compiler will report an error that must be fixed before the code will compile. There are many other examples but I don't see a reason to document them all here.

Note I say "compile" and not "compile to a working binary" because the more typical result is the compiler correctly building the input which then generates a SIGILL when executed. This is covered by the "Not all such issues..." sentence.  Personally I don't think it's worth going into anymore detail than this.

mgabka: Sure, I get the difference between "compile" and not "compile to a working binary". 

For me it was just the meaning of word support, as in my mind feature could be supported (as implemented) by a tool but not available when given set of options it provided. It has this small difference from the user point of view as if feature is not implemented they can not do anything, while if it is not enabled they can change options.
But I am not against your version either.

paulwalker-arm: I think that's because you are considering this from the point of view of a human asking whether the compiler can support a feature. Whereas the feature macros are source code actions and thus the context is the point the source code is being compiled, at which point the compiler has already been "configured" (for want of a better word) by the user.

Perhaps the point you are raising is the document does not detail what actions the user should take in order to control the definitions of the feature flags, which is intentional because we cannot dictate that.

rsandifo-arm: I agree it would be good to be a bit more precise about what “support” means though.  At one level, GCC 14 “supports” SME2, and the word is often used in that sense.  So if we just say “support”, users might expect GCC 14 to define `__ARM_FEATURE_SME2` unconditionally.

Instead, for compatibility, we have to define the macro only if the intrinsics are “available” in functions that have no target annotations (not precise wording either).  What users need to do to achieve that for a given compiler is outside the scope of the ACLE, but I think the result is within scope.

IMO, the example of using SME intrinsics in non-streaming functions is an example of an intrinsics being available, but being used incorrectly.  It's somewhat like passing an out-of-range value to a constant argument, except that the error is contextual, rather than contained within the call.  (I still think it's a good example though.)

CarolineConcatto:
Thank you @paulwalker-arm, @mgabka  and @rsandifo-arm  for your inputs. 
I have made the changes asked.
Carol

main/acle.md
CarolineConcatto: I did changed to what @paulwalker-arm requested, but instead of using support I wrote available, as suggested by @rsandifo-arm.
I hope it is correct.

DanielKristofKiss (approved):
LGTM

vhscampos:
Let's wait until the end of this week for additional feedback

sdesmalen-arm (changes requested):

main/acle.md
```diff
@@ -925,8 +925,9 @@ and:
 to the more specific header files below. These intrinsics are in the
 C implementation namespace and begin with double underscores. It is
 unspecified whether they are available without the header being
-included. The `__ARM_ACLE` macro should be tested before including the
-header:
+included. The `__ARM_ACLE` macro can be tested before including the header.
+However, this is not recommended for cases where target feature selection is
+overridden within code, for example, when using #function-multi-versioning.
```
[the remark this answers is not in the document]
rsandifo-arm: Agreed, but also: `__ARM_ACLE` is a special case in that, like `__ARM_NEON_SVE_BRIDGE`, it describes the presence of a header file rather than a specific architecture feature.  So I think the original text was ok too (and so is your suggestion).

For the feature macros, maybe “is available in a normal unannotated function” would be more specific than “is available”.

main/acle.md
```diff
@@ -1072,6 +1083,31 @@ including the header:
 
 Including `<arm_sme.h>` also includes [`<arm_sve.h>`](#arm_sve.h).
 
+### Predefined feature macros and header files
+
+Testing a feature macro returns the compiler's availability for the feature via ACLE
+documented intrinsics and inline assembly. This property can be independent of
+where the test is performed, meaning a feature macro does not guarantee the
+validity of using the feature at the point it is tested. For example:
```
rsandifo-arm, line 1091:
```suggestion
Feature macros indicate whether that feature's intrinsics and inline assembly instructions
are available in normal unannotated functions. In some implementations, the macro definitions
are invariant throughout a translation unit, while in other implementations, the macro
definitions can vary with pragmas.

The macro definitions are otherwise independent of where the test is performed,
meaning that a feature macro does not guarantee that the feature can be used at the
point that it is tested. For example:
```

main/acle.md
```diff
@@ -1072,6 +1083,31 @@ including the header:
...
+    #ifdef __ARM_FEATURE_SME
+      svst1_hor_za8(0, slice_base, pg, ptr);
+    #endif
+    }
+```
+
+Whilst testing `__ARM_FEATURE_SME` ensures the compiler has available the SME
+intrinsic `svst1_hor_za8`, `foo` will fail to compile because it has not been
+decorated with the necessary keywords to guarantee the function will be
+executed in streaming mode.
```
rsandifo-arm, line 1106:
```suggestion
While testing `__ARM_FEATURE_SME` ensures that the SME intrinsic `svst1_hor_za8`
is available, `foo` will fail to compile because the call does not occur in a
[streaming statement](#streaming-statement).
```

main/acle.md
```diff
@@ -1072,6 +1083,31 @@ including the header:
...
+
+Not all such issues can be caught during compilation and may instead result in
+runtime failures.
```
rsandifo-arm, line 1109: Do you know of a specific example where that's true?

main/acle.md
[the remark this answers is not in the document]
CarolineConcatto: So this was suggested by @paulwalker-arm 

The `XYZ` macro can be tested before including the header. However, this is not recommended for cases where target feature selection is overridden within code, for example, when using #function-multi-versioning.
From another previous asked change.
 
I have no preference, but would be nice to agree in what to write here, before I change again.  The sentence you see here is a mix of previous interaction that we agree we should replace taking into account FMV.

main/acle.md
```diff
@@ -1072,6 +1083,31 @@ including the header:
 
 Including `<arm_sme.h>` also includes [`<arm_sve.h>`](#arm_sve.h).
 
+### Predefined feature macros and header files
+
```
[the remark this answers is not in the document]
CarolineConcatto: The goal here is to show that guarding  code with macros does not guarantee  a working binary, because this intrinsics may need more than that.

main/acle.md
CarolineConcatto: I cannot think of anything that we have now.
The frontend is doing a decent job protecting the intrinsics at compiler time. 
I can remove if you think is best

main/acle.md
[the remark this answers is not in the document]
rsandifo-arm: I'm not sure that wording is appropriate for `__ARM_ACLE` though (specifically that macro, not others).  There is no change in features that would affect whether the compiler provides `<arm_acle.h>`.  Intrinsics within `<arm_acle.h>` that depend on features have their own `__ARM_FEATURE_*` macro.

main/acle.md
rsandifo-arm: Yeah, think it would be best to remove it if we can't give a specific example.

main/acle.md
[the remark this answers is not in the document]
sdesmalen-arm: The current description is better, but I wonder if it would be sufficient just to say: "No assumptions should be made on the order or context in which the preprocessor macros are evaluated, for example:"

What do you think @CarolineConcatto @rsandifo-arm  ?

main/acle.md
```diff
@@ -1072,6 +1083,34 @@ including the header:
...
+meaning that a feature macro does not guarantee that the feature can be used at the
+point that it is tested. For example:
+
+``` c
+    #include <arm_sme.h>
+
+    void foo(svbool_t pg, void *ptr, uint32_t slice_base) {
+    #ifdef __ARM_FEATURE_SME
+      svst1_hor_za8(0, slice_base, pg, ptr);
+    #endif
```
sdesmalen-arm, line 1103: I thought the paragraph above is more about the evaluation context/order of the macro to express that a user can't rely on the macro being set/unset when used with the ACLE keyword attributes, e.g.

```
__attribute__((target("+sve")))
void foo() {
#ifdef __ARM_FEATURE_SVE
  // The user should make no assumptions that the target attribute
  // has enabled the __ARM_FEATURE_SVE macro.
#endif
}
```

The current example is more about showing that the compiler can add additional restrictions to the use of an intrinsic beyond what is captured by the ACLE macro. If that is the intent of the example, I think it would be helpful if that can be added as an separate paragraph, e.g. something like "The compiler may add additional restrictions to the intrinsics beyond what is captured by the ACLE macros depending on the context in which the intrinsics are used".

CarolineConcatto: Thank you @sdesmalen-arm  for the suggestions.
I made some changes according to your comments.

main/acle.md
```diff
@@ -939,8 +939,8 @@ header:
 `<arm_fp16.h>` is provided to define the scalar 16-bit floating point
 arithmetic intrinsics. As these intrinsics are in the user namespace,
 an implementation would not normally define them until the header is
-included. The `__ARM_FEATURE_FP16_SCALAR_ARITHMETIC` feature macro
-should be tested before including the header:
+included. When `__ARM_FEATURE_FP16_SCALAR_ARITHMETIC` is defined to `1`,
+the header file is available in a normal unannotated function.
```
sdesmalen-arm, line 943: The macro is also available outside of function scope. The scope in which it is used shouldn't matter, because it is a preprocessor macro.

main/acle.md
```diff
@@ -1072,6 +1073,30 @@ including the header:
 
 Including `<arm_sme.h>` also includes [`<arm_sve.h>`](#arm_sve.h).
 
+### Predefined feature macros and header files
+
+Testing a feature macro returns the compiler's availability for the feature
+via ACLE documented intrinsics and inline assembly. This property can be
+independent of where the test is performed. There should be no assumption on
+the order or context in wich the preprocessor macros are evaluated.
+
```
sdesmalen-arm, line 1082: I think it's worth including the other example I shared here.

main/acle.md
```diff
@@ -1072,6 +1073,30 @@ including the header:
 
 Including `<arm_sme.h>` also includes [`<arm_sve.h>`](#arm_sve.h).
 
+### Predefined feature macros and header files
+
+Testing a feature macro returns the compiler's availability for the feature
+via ACLE documented intrinsics and inline assembly. This property can be
+independent of where the test is performed. There should be no assumption on
+the order or context in wich the preprocessor macros are evaluated.
```
sdesmalen-arm, line 1081: I find the word 'test' confusing, would it make more sense to speak of 'evaluating' ?

```suggestion
Evaluating a feature macro returns the availability of intrinsics and inline
assembly for that feature, but no assumptions should be made on the order or
context in which the preprocessor macros are evaluated.
```

main/acle.md
```diff
@@ -1072,6 +1073,30 @@ including the header:
...
+    void foo(svbool_t pg, void *ptr, uint32_t slice_base) {
+    #ifdef __ARM_FEATURE_SME
+      svst1_hor_za8(0, slice_base, pg, ptr);
+    #endif
+    }
+```
+
+While testing `__ARM_FEATURE_SME` ensures that the SME intrinsic `svst1_hor_za8`
+is available, `foo` will fail to compile because the call does not occur in a
+[streaming statement](#streaming-statement).
```
sdesmalen-arm, line 1098:
```suggestion
If `__ARM_FEATURE_SME` evaluates to `true` the SME intrinsic `svst1_hor_za8`
is available, but `foo` may still fail to compile because the call does not occur in a
[streaming statement](#streaming-statement).
```

main/acle.md
CarolineConcatto: I wrote again this sentence to:
When `__ARM_xxx` is defined to `1`, the header file is available
regardless of the context in which the macro is evaluated.

main/acle.md
```diff
@@ -1019,7 +1023,7 @@ Including `<arm_sve.h>` also includes the following header files:
 
 `<arm_neon_sve_bridge.h>` defines intrinsics for moving data between
 Neon and SVE vector types; see [NEON-SVE Bridge](#neon-sve-bridge)
-for details.  The `__ARM_NEON_SVE_BRIDGE` macro should be tested
+for details. The `__ARM_NEON_SVE_BRIDGE` macro should be tested
```
sdesmalen-arm, line 1026: nit: unnecessary change.

main/acle.md
```diff
@@ -1072,6 +1076,39 @@ including the header:
...
+Evaluating a feature macro returns the availability of intrinsics and inline
+assembly for that feature, but no assumptions should be made on the order or
+context in which the preprocessor macros are evaluated. For example:
+
+``` c
+    __attribute__((target("+sve")))
+    void foo() {
+    #ifdef __ARM_FEATURE_SVE
+      // The user should make no assumptions that the target attribute
+     // has enabled the __ARM_FEATURE_SVE macro.
```
sdesmalen-arm, line 1090: nit: whitespace
```suggestion
      // The user should make no assumptions that the target attribute
      // has enabled the __ARM_FEATURE_SVE macro.
```

main/acle.md
```diff
@@ -1072,6 +1076,39 @@ including the header:
...
+    #endif
+}
+```
+
+The compiler may add additional restrictions to the intrinsics beyond what is
+captured by the ACLE macros depending on the context in which the intrinsics
+are used. For example:
```
sdesmalen-arm, line 1097: I'm still not convinced this is worth pointing out, because the compiler may also enforce requirements on the range of immediate arguments, and we never bothered to document that either. That said though, I don't see any harm in adding this either.

sdesmalen-arm (approved):
I'm content with the phrasing now. Please address the nits though.

vhscampos:
Since this PR got approval with minor nits, I'm merging it. @CarolineConcatto please address the remaining nits in a following PR. Thanks

main/acle.md
```diff
@@ -1072,6 +1076,39 @@ including the header:
...
+    #include <arm_sme.h>
+    void foo(svbool_t pg, void *ptr, uint32_t slice_base) {
+    #ifdef __ARM_FEATURE_SME
+      svst1_hor_za8(0, slice_base, pg, ptr);
+    #endif
+    }
+```
+
+If `__ARM_FEATURE_SME` evaluates to `true` the SME intrinsic `svst1_hor_za8`
+is available, but `foo` may still fail to compile because the call does not
```
sallyarmneale, line 1109: may -> might
https://confluence.arm.com/display/TCOM/Concise+and+clear+words+and+phrases

sallyarmneale (commented):
Two minor changes

PR #331, opened 2024-07-22
Replace latin contractions

vhscampos:
Replace latin contractions in accordance to Arm's writing guidelines.


sallyarmneale (commented):
Looks good to me.

PR #333, opened 2024-07-22
Limit the run of draftversion-check to the relevant branches

vhscampos:
The draftversion checker must run only in the main branch. This patch limits the run conditions to abide by this.


vhscampos:
Self-approved: maintenance patch

PR #336, opened 2024-07-26
[SME] Add __arm_agnostic("sme_za_state") keyword attribute

sdesmalen-arm:
The `__arm_agnostic` keyword attribute enables the user to specify that a function is agnostic to a specified piece of architectural state. That means that the function must preserve this state when it exists, or otherwise ignores its contents.

The reason for not naming this something like `__arm_za_compatible` was so that we might want use the attribute keyword for other architectural state in the future.

---
name: Pull request

---

If this PR is about a bugfix:

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.


sdesmalen-arm:
@rsandifo-arm

main/acle.md
```diff
@@ -832,6 +833,7 @@ predefine the associated macro to a nonzero value.
 | [`__arm_new`](#arm_new)                                     | function declaration  | Argument-dependent                |
 | [`__arm_out`](#ways-of-sharing-state)                       | function type         | Argument-dependent                |
 | [`__arm_preserves`](#ways-of-sharing-state)                 | function type         | Argument-dependent                |
+| [`__arm_agnostic`](#arm_agnostic)                           | function type         | `__ARM_FEATURE_SME`               |
```
rsandifo-arm, line 836: Nit: would be good to keep the entries sorted by name.

main/acle.md
```diff
@@ -4836,6 +4838,47 @@ if such a restoration is necessary. For example:
...
+
+*   If PSTATE.ZA is available, then the [abstract machine](#abstract-machine)
+    ensures that on return from the function, the value of PSTATE.ZA is the same
+    as it was on entry to the function.
+
+*   If the function forms part of the object code's ABI, that object code
+    function has a “ZA-compatible interface”; see [[AAPCS64]](#AAPCS64)
+    for more details.
+
+*   It is not valid for a function declaration with `__arm_agnostic("sme_za_state")`
```
rsandifo-arm, line 4871: `__arm_new` is only meaningful for definitions, not declarations.  I'm not sure it should be a semantic requirement, since:
```
__arm_new("za") void foo(void) …stuff… {
    …
}
```
isn't semantically different from:
```
__arm_new("za") void foo1(void) {
    …
}
void foo(void) …stuff… {
    foo1();
}
```
(It's reasonable for an implementation to say that it doesn't support `__arm_new` for now though.  This would be the equivalent of a GCC “sorry” diagnostic, but it'd be an implementation choice rather than something that the ACLE should document.)

I think we can then say:

> `It is not valid … to [share](#shares-state) PSTATE.ZA state with its caller.`

main/acle.md
```diff
@@ -4836,6 +4838,47 @@ if such a restoration is necessary. For example:
...
+preserve the architectural state that is specified by its arguments when such
+state exists at runtime. The function is otherwise unconcerned with the contents
+of this state.
+
+The `__arm_agnostic` [keyword attribute](#keyword-attributes) applies to
+**function types** and accepts the following arguments:
+
+```"sme_za_state"```
+
+*   If the function is defined and PSTATE.ZA is available, the definition must
```
rsandifo-arm, line 4853: Since the AAPCS64 patch (rightly) creates a three-way choice between private, shared & agnostic, we should update the description under “Mapping to the Procedure Call Standard” section.  I think the text here should then say something like:

> This attribute affects the ABI of object code functions, as described under …  It is the compiler's responsibility to ensure that such object code functions honor the ABI requirements.

and then delete the following two bullet points.  AFAICT, the attribute has no effect on the behaviour of the abstract machine outside of inline asms.

rsandifo-arm (commented):
The patch should also update the “Inline assembly” section, since ZA can now be “on” on entry to an inline asm in cases that the current text doesn't cover.

main/acle.md
```diff
@@ -4855,7 +4885,17 @@ one of the following is true:
 
 *   F shares ZT0 with its caller
 
-All other functions have a private-ZA interface.
+All other functions have either a private-ZA or an agnostic-ZA interface.
```
rsandifo-arm, line 4888: I think we should leave this sentence as-is and describe the agnostic-ZA case before it.  I'd recomment restructuring the text as follows:

```
If a C or C++ function F forms part of the object code's ABI:

*   the object code function has a shared-ZA interface if and only if at least
    one of the following is true:

    *   F shares ZA with its caller

    *   F shares ZT0 with its caller

*   the object code function has an agnostic-ZA interface if and only if
    F's type has an `__arm_agnostic("sme_za_state")` attribute.

All other functions have either a private-ZA interface.
```

main/acle.md
```diff
@@ -4847,6 +4873,10 @@ interfaces:
 
 *   a “shared-ZA” interface
 
+<span id="agnostic-za"></span>
+
+*   a "agnostic-ZA" interface
```
rsandifo-arm, line 4878:
```suggestion
*   an "agnostic-ZA" interface
```

main/acle.md
```diff
@@ -4855,7 +4885,17 @@ one of the following is true:
...
-All other functions have a private-ZA interface.
+All other functions have either a private-ZA or an agnostic-ZA interface.
+
+If F implements an agnostic-ZA interface and PSTATE.ZA is available at runtime,
+then a call to F must return with its ZA state unchanged in accordance
+to the [[AAPCS64]](#AAPCS64). In practice this means that calls to F don't have
+to emit code to set up a lazy-save for ZA or to preserve other state like ZT0
+when such state is live at the call site.
+
+The implementation of F must not make any assumptions on the availability of
```
rsandifo-arm, line 4896: I think we should instead add a note to the attribute documentation to say that it does not imply that SME is available.

rsandifo-arm (commented):
The previous comment about updating the inline asm documentation still stands.

sdesmalen-arm:
> The previous comment about updating the inline asm documentation still stands.

Thanks, I missed that comment previously. I've now added a comment that I believe is sufficient to cover the agnostic-ZA case, but let me know if there's still anything missing!

main/acle.md
```diff
@@ -4847,16 +4874,34 @@ interfaces:
...
+  * F shares ZA with its caller
+
+  * F shares ZT0 with its caller
+
+* the object code function has an agnostic-ZA interface if and only if F's type
+  has an `__arm_agnostic("sme_za_state")` attribute.
 
 All other functions have a private-ZA interface.
 
+If F implements an agnostic-ZA interface and ZA is active, then a call to F must
```
rsandifo-arm, line 4895: I still think we should drop these two paragraphs.  IMO the purpose of this section is to describe how ACLE features map to the ABI, and it's the ABI's job to describe the actual binary interface.

I was hoping the new “The use of this attribute does not imply that SME is available.” statement would replace the second paragraph, since IMO it's more precise.  E.g. it's valid to do:

```
  void f() __arm_agnostic("sme_state") __arm_streaming {}
```

and in that case, the implementation _can_ assume that PSTATE.ZA is available.

In other words, the point is that `__arm_agnostic` does not _by itself_ guarantee that SME is available, but the compiler might still know that SME is available from other sources, e.g. from `+sme` or from other attributes.  `__arm_agnostic` doesn't override information from other sources.

main/acle.md
```diff
@@ -4942,6 +4987,8 @@ following is true:
 
 *   F [uses](#uses-state) `"zt0"`
 
+*   F implements an [agnostic ZA](#agnostic-za) interface.
```
rsandifo-arm, line 4990: Looks good, but this should be reflected in the table at the end of the section as well.

I think we also need to decide what to do about:

```
*   S is not an input to the asm and S is not an output from the asm.
    However, the contents of S after executing the asm's instructions might
    be different from the contents of S before executing the instructions.
    This is indicated by adding S's string to the asm's clobber list.

    If an asm takes this option for state that is controlled by PSTATE.ZA,
    the asm itself is responsible for handling the [[AAPCS64]](#AAPCS64)
    lazy save scheme.
```

The current rule implicity assumes that ZA can only be dormant or off.  If ZA is active, then is the asm required to preserve ZA despite the clobber?  Does the compiler do this?  (In which case, does it also handle the lazy-save scheme, meaning that the asm doesn't need to?)  Or is this option not valid in agnostic-ZA functions?

sdesmalen-arm: My thinking on this was that for `__arm_agnostic("sme_za_state")` functions, if the asm clobbers any of the state enabled by PSTATE.ZA, the user is responsible for letting the function preserve this state using the provided routines such that on return from the function the ABI is honoured.

main/acle.md
```diff
@@ -4983,14 +4974,22 @@ function F if at least one of the following is true:
 Otherwise, ZA can be in any state on entry to A if at least one of the
 following is true:
 
-*   F [uses](#uses-state) `"za"`
+*   F [uses](#uses-state) `"za"`.
+
+*   F [uses](#uses-state) `"zt0"`.
+
+*   F is what the [[AAPCS64]](#AAPCS64) calls an ["agnostic-ZA"](#agnostic-za)
+    function and A's clobber-list does not include `"za"` or `"zt0"`.
```
rsandifo-arm, line 4982:
```suggestion
*   F's type has an [`__arm_agnostic("sme_za_state")`  attribute](#agnostic-za)
    and A's clobber-list includes neither `"za"` nor `"zt0"`.
```
so that we keep it in terms of C/C++-level constructs.

main/acle.md
```diff
@@ -4983,14 +4974,22 @@ function F if at least one of the following is true:
...
+
+Otherwise, ZA can be off or dormant on entry to A if at least one of the
```
rsandifo-arm, line 4984: I think we should just keep the existing “Otherwise…” paragraph as-is.  AFAIK, there are no “otherwise” cases in which ZA is guaranteed to be off on entry.

main/acle.md
```diff
@@ -5015,9 +5014,16 @@ depend on ZT0 as well as ZA.
...
-| dormant               | dormant              | " " "                                  |
-| dormant               | off                  | " " ", and A clobbers `"za"`           |
-| active                | active               | F uses `"za"` and/or `"zt0"`           |
+|                       |                      | or F is an                             |
+|                       |                      | ["#agnostic-ZA"](#agnostic-za)         |
+|                       |                      | function.                              |
+| dormant               | dormant              | "" "" ""                               |
+| dormant               | off                  | A clobbers `"za"`, but F does not      |
+|                       |                      | [use](#uses-state) `"za"`or `"zt0"`.   |
+| active                | active               | F uses `"za"` and/or `"zt0"`, or       |
```
rsandifo-arm, line 5023: I think it's only this row that needs to change.  The old version of the others is still correct.

I'd suggest the same “neither…nor” change as above, but it's hard to get the suggestion right in the github editor :-)

rsandifo-arm (approved):
LGTM.

main/acle.md
```diff
@@ -5070,13 +5097,21 @@ interfaces:
...
-*   F shares ZA with its caller
+* the object code function has a shared-ZA interface if and only if at least
+  one of the following is true:
 
-*   F shares ZT0 with its caller
+  * F shares ZA with its caller
+
+  * F shares ZT0 with its caller
+
+* the object code function has an agnostic-ZA interface if and only if F's type
```
philgee-oss, line 5113: interface only if the type for F

main/acle.md
```diff
@@ -5070,13 +5097,21 @@ interfaces:
...
-object code function has a shared-ZA interface if and only if at least
-one of the following is true:
+<span id="agnostic-za"></span>
+
+*   an "agnostic-ZA" interface
+
+If a C or C++ function F forms part of the object code's ABI:
 
-*   F shares ZA with its caller
+* the object code function has a shared-ZA interface if and only if at least
```
philgee-oss, line 5106: interface only if

main/acle.md
```diff
@@ -5059,6 +5061,31 @@ if such a restoration is necessary. For example:
...
+such state exists at runtime. The function is otherwise unconcerned with this
+state.
+
+The `__arm_agnostic` [keyword attribute](#keyword-attributes) applies to
+**function types** and accepts the following arguments:
+
+```"sme_za_state"```
+
+*   This attribute affects the ABI of a function, which must implement an
+    [agnostic-ZA interface](#agnostic-za). It is the compiler's responsibility
```
philgee-oss, line 5077: It is the responsibility of the compiler
    to ensure that the object code of the function honors the ABI requirements.

PR #346, opened 2024-09-04
[FMV] Unify ls64, ls64_v and ls64_accdata.

labrinea:
As originally discussed in #315 and then in #329 we want to unify features that the toolchains cannot support independently. In the case of ls64 I have attempted to split the lumped feature in the compiler (see https://github.com/llvm/llvm-project/pull/101712), but unfortunately this would break backwards compatibility:

Mapping 'ls64' to FeatureLS64_ACCDATA would enable all three of {FeatureLS64, FeatureLS64_V, FeatureLS64_ACCDATA} but then using 'nols64' either on the command line or the assembler directive would only disable FeatureLS64_ACCDATA without disabling the other two. For that we would have to map 'ls64' to FeatureLS64, but then it would not enable the other two.

As far as I am aware there are no hardware implementations out there which implement ls64 without ls64_v or ls64_accdata, so unifying these features in the specification should not be a regression for feature detection. If any of this changes in the feature we can revisit the specification.

---
name: Pull request

---

If this PR is about a bugfix:

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -403,6 +403,7 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * Added `__FUNCTION_MULTI_VERSIONING_SUPPORT_LEVEL` to indicate the support
   level of the [Function Multi Versioning](#function-multi-versioning).
 * Unified Function Multi Versioning features sha1, sha2.
+* Unified Function Multi Versioning features ls64, ls64_v, ls64_accdata.
 * Added [**Alpha**](#current-status-and-anticipated-changes)
   support for SME2.1 (FEAT_SME2p1).
 * Removed requirement to add preprocessor guards for header files.
@@ -2755,9 +2756,7 @@ The following table lists the architectures feature mapping for AArch64
    | 490           | `FEAT_SSBS`              | ssbs          | ```ID_AA64PFR1_EL1.SSBS >= 0b0001```      |
    | 500           | `FEAT_SSBS2`             | ssbs2         | ```ID_AA64PFR1_EL1.SSBS >= 0b0010```      |
    | 510           | `FEAT_BTI`               | bti           | ```ID_AA64PFR1_EL1.BT >= 0b0001```        |
-   | 520           | `FEAT_LS64`              | ls64          | ```ID_AA64ISAR1_EL1.LS64 >= 0b0001```     |
-   | 530           | `FEAT_LS64_V`            | ls64_v        | ```ID_AA64ISAR1_EL1.LS64 >= 0b0010```     |
-   | 540           | `FEAT_LS64_ACCDATA`      | ls64_accdata  | ```ID_AA64ISAR1_EL1.LS64 >= 0b0011```     |
+   | 520           | `FEAT_LS64`, `FEAT_LS64_V`, <br> `FEAT_LS64_ACCDATA` | ls64 | ```ID_AA64ISAR1_EL1.LS64 >= 0b0011``` |
    | 550           | `FEAT_WFxT`              | wfxt          | ```ID_AA64ISAR2_EL1.WFxT >= 0b0010```     |
    | 560           | `FEAT_SME_F64F64`        | sme-f64f64    | ```ID_AA64SMFR0_EL1.F64F64 == 0b1```      |
    | 570           | `FEAT_SME_I16I64`        | sme-i16i64    | ```ID_AA64SMFR0_EL1.I16I64 == 0b1111```   |


labrinea:
@DanielKristofKiss @jroelofs @tmatheson-arm @andrewcarlotti

main/acle.md
```diff
@@ -2756,9 +2757,7 @@ The following table lists the architectures feature mapping for AArch64
    | 490           | `FEAT_SSBS`              | ssbs          | ```ID_AA64PFR1_EL1.SSBS >= 0b0001```      |
    | 500           | `FEAT_SSBS2`             | ssbs2         | ```ID_AA64PFR1_EL1.SSBS >= 0b0010```      |
    | 510           | `FEAT_BTI`               | bti           | ```ID_AA64PFR1_EL1.BT >= 0b0001```        |
-   | 520           | `FEAT_LS64`              | ls64          | ```ID_AA64ISAR1_EL1.LS64 >= 0b0001```     |
-   | 530           | `FEAT_LS64_V`            | ls64_v        | ```ID_AA64ISAR1_EL1.LS64 >= 0b0010```     |
-   | 540           | `FEAT_LS64_ACCDATA`      | ls64_accdata  | ```ID_AA64ISAR1_EL1.LS64 >= 0b0011```     |
+   | 520           | `FEAT_LS64`,`FEAT_LS64_V`,<br>`FEAT_LS64_ACCDATA`| ls64 | ```ID_AA64ISAR1_EL1.LS64 >= 0b0011``` |
```
tmatheson-arm, line 2760:
```suggestion
   | 520           | `FEAT_LS64_ACCDATA`| ls64 | ```ID_AA64ISAR1_EL1.LS64 >= 0b0011``` |
```
I don't think there should be multiple architecture features listed. It makes it ambiguous how you test if the feature is present or not.

labrinea: That's the whole point I am trying to make, I am suggesting that the FMV feature named `ls64` means all of {FEAT_LS64,FEAT_LS64_V,FEAT_LS64_ACCDATA}. It is better to make it explicit in the specification rather than letting the user figure out the architecture dependency FEAT_LS64 -> FEAT_LS64_V ->FEAT_LS64_ACCDATA themselves in my opinion.

DanielKristofKiss: Relevant features for others items are listed too in the table already.
Here this is a favour for the reader how looks for `FEAT_LS64` to be sure the `ls64` is the right option to choose without looking up the bits in the `ID_AA64ISAR1_EL1`

labrinea: As far as I understand we don't want to list architectural features that are implied if they have a corresponding FMV feature in place.

tmatheson-arm: Ok fair enough

tmatheson-arm, line 2760:
```suggestion
   | 520           | `FEAT_LS64`,`FEAT_LS64_V`,<br>`FEAT_LS64_ACCDATA`| ls64_accdata | ```ID_AA64ISAR1_EL1.LS64 >= 0b0011``` |
```
I'm not sure which way to go here but it's worth discussing. Generally FMV keeps the names of the features matching with the architectural `FEAT_` names, and the current PR makes `ls64` inconsistent with `FEAT_LS64_ACCDATA`. If we ever did want to add independent `FEAT_LS64` and `FEAT_LS64_V` to FMV, we would have the same problem we have in the compilers where `ls64` can't be reassigned from `FEAT_LS64_ACCDATA` to `FEAT_LS64`. However I can see that it makes the feature name inconsistent with the existing compiler command line spellings.

tmatheson-arm: Being inconsistent with the command line flags could be mitigated with a warning like `no such feature ls64; did you mean ls64_accdata?`

labrinea: I think we want to stay aligned with the command line/assembler directive here, otherwise we will create further confusion. But that's my opinion. @andrewcarlotti, any opinion?

andrewcarlotti: I agree that FMV feature names should match command line and assembler directive names.

labrinea:
> LGTM. If we do find a convincing argument for splitting the feature then we'll have to revisit this, but a split would need to use different names to the ones proposed in the original (current) spec, for consistency with existing established command line options.

But how about this issue I mentioned in the description?

> Mapping 'ls64' to FeatureLS64_ACCDATA would enable all three of {FeatureLS64, FeatureLS64_V, FeatureLS64_ACCDATA} but then using 'nols64' either on the command line or the assembler directive would only disable FeatureLS64_ACCDATA without disabling the other two. For that we would have to map 'ls64' to FeatureLS64, but then it would not enable the other two.

How are we going to deal with this in the future if we ever decide to split the feature?

andrewcarlotti:
> > LGTM. If we do find a convincing argument for splitting the feature then we'll have to revisit this, but a split would need to use different names to the ones proposed in the original (current) spec, for consistency with existing established command line options.
> 
> But how about this issue I mentioned in the description?
> 
> > Mapping 'ls64' to FeatureLS64_ACCDATA would enable all three of {FeatureLS64, FeatureLS64_V, FeatureLS64_ACCDATA} but then using 'nols64' either on the command line or the assembler directive would only disable FeatureLS64_ACCDATA without disabling the other two. For that we would have to map 'ls64' to FeatureLS64, but then it would not enable the other two.
> 
> How are we going to deal with this in the future if we ever decide to split the feature?

We can worry about that issue in the future if necessary; it doesn't affect the decisions we need to make now. 

(However, one possible solution could be to create three new flags, and the existing flag would become an umbrella flag that enables or disables all of them (similarly to `crypto`, although with less quirks and inconsistencies)).

labrinea:
Not direclty an ACLE question, but it relates to this patch. In the CPUFeatures enum which name shall be used to indicate all three {FeatureLS64, FeatureLS64_V, FeatureLS64_ACCDATA}? FEAT_LS64 or FeatureLS64_ACCDATA? For other entries with more than one features listed in ACLE we use the top level dependency. For example on {FEAT_SM3,FEAT_SM4} we use FEAT_SM4. On that ground I would guess FeatureLS64_ACCDATA is the one to keep?

tmatheson-arm:
> In the CPUFeatures enum which name shall be used to indicate all three

I would say ls64_accdata, but it's an implementation detail (the name is not part of the ABI, only the bit position)

vhscampos:
@labrinea The PDF rendering is broken in the table line with priority 520. Can you please fix it?
You can find the output PDF at the bottom of this page: https://github.com/ARM-software/acle/actions/runs/10703762051

vhscampos (changes requested):
Broken rendering

vhscampos (approved):
LGTM. Thanks for the patch

main/acle.md
```diff
@@ -403,6 +403,7 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * Added `__FUNCTION_MULTI_VERSIONING_SUPPORT_LEVEL` to indicate the support
   level of the [Function Multi Versioning](#function-multi-versioning).
 * Unified Function Multi Versioning features sha1, sha2.
+* Unified Function Multi Versioning features ls64, ls64_v, ls64_accdata.
```
sallyarmneale, line 406: missing AND in the list:
ls64, ls64_v, and ls64_accdata.

main/acle.md
```diff
@@ -403,6 +403,7 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * Added `__FUNCTION_MULTI_VERSIONING_SUPPORT_LEVEL` to indicate the support
   level of the [Function Multi Versioning](#function-multi-versioning).
 * Unified Function Multi Versioning features sha1, sha2.
```
sallyarmneale, line 405: sha1 and sha2.

sallyarmneale (commented):
2 tiny changes.
List of 2 - needs an AND not a comma
List of 3, use Oxford/Serial Comma, see https://confluence.arm.com/pages/viewpage.action?pageId=1589416839

PR #351, opened 2024-09-27
[FMV] Unify memtag and memtag2.

labrinea:
If we split these features in the compiler (see relevant pull request https://github.com/llvm/llvm-project/pull/109299), we would only be able to hand-write a 'memtag2' version using inline assembly since the compiler cannot generate the instructions that become available with FEAT_MTE2. On top of that these instructions only work at Exception Level 1, so they would be unusable since FMV is a user space facility. I am therefore unifying them in the ACLE specification.

---
name: Pull request

---

If this PR is about a bugfix:

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -416,6 +416,7 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 #### Changes for next release
 
 * Unified Function Multi Versioning features ssbs and ssbs2.
+* Unified Function Multi Versioning features memtag and memtag2.
 
 ### References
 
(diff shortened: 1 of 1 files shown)


labrinea:
@vhscampos @andrewcarlotti @tmatheson-arm

andrewcarlotti:
Ok, here's a suggestion that might seem to go against stuff I've previously said. How about we call the FMV feature "memtag2" to deliberately distinguish it from the "memtag" command line option? The justification for this is:
- As a command line feature, "memtag" is enabled for cpus that might only support FEAT_MTE (and not FEAT_MTE2, depending upon configuration).
- If we're using HWCAPS to control FMV feature gating, then the available HWCAP corresponds to FEAT_MTE2.
- If we don't have FEAT_MTE2 at runtime, then the only benefit to using FEAT_MTE instructions would be to avoid runtime dispatch. But using FMV implies that we're already doing runtime dispatch, so there's no benefit to having FMV enable usage of the FEAT_MTE instructions on a system without FEAT_MTE2.

So the FMV feature "memtag2" would deliberately have a slightly different definition to the command line option "memtag". Internally the compiler may choose to handle this by mapping the FMV feature directly to the command line option, with the caveat that enabling the command line option does not necessarily imply that the FMV feature will be enabled at runtime.

(Adding @Wilco1 as well)

tmatheson-arm:
My understanding of FEAT_MTE vs FEAT_MTE2 is that (excepting the small number of EL1 instructions that are FEAT_MTE2 only) they offer the same set of instructions, which
- on a system with FEAT_MTE2, perform tagging operations and checks
- on a system with only FEAT_MTE, are effectively NOPS
- on a system with neither, are undefined.

As far as the compiler is concerned, there is no difference between FEAT_MTE and FEAT_MTE2. The codegen should be identical. This is probably why they were historically combined into `+memtag`. This means that you can compile with `+memtag`, and if you have FEAT_MTE, your code will not crash but you won't get any checking, and with FEAT_MTE2 you will get checking. This is also why you can usually configure FEAT_MTE2 at boot, but FEAT_MTE is always on.

FMV complicates the situation because it is for runtime dispatch, and there is a difference between systems that implement FEAT_MTE and FEAT_MTE2. This is why it introduced `+memtag` and `+memtag2`. The codegen should be the same, but you can do dispatch on them. It's not clear whether that's important.

You might want to detect whether you have FEAT_MTE2 rather than FEAT_MTE in order to use the EL1 instructions. However, apparently [FMV targets EL0](https://github.com/llvm/llvm-project/pull/109299#issuecomment-2373845701) which would mean there is no need to support that in FMV.

Since we have gone down the path of keeping FMV names in sync with command line names, I think it would be silly to deviate from that at this point. I suggest we keep `memtag` == FEAT_MTE2.

labrinea:
> If we're using HWCAPS to control FMV feature gating, then the available HWCAP corresponds to FEAT_MTE2.

I find this argument irrelevant. There are other features we intend to detect with HWCAPS which are available to the user under a different name than the HWCAP macro (HWCAP_PMULL detects "aes" for example), or features whose name corresponds to a different architectural extension (HWCAP_SSBS detects "ssbs" which means FEAT_SSBS2).

> If we don't have FEAT_MTE2 at runtime, then the only benefit to using FEAT_MTE instructions would be to avoid runtime dispatch.

Can you explain what do you mean here? I am not getting it.

> Since we have gone down the path of keeping FMV names in sync with command line names, I think it would be silly to deviate from that at this point. I suggest we keep memtag == FEAT_MTE2.

I agree.

DanielKristofKiss (approved):
LGTM.
for the developers memtag here would mean tagging is available and usable.
maybe memtag2 could be an alias to memtag.

PR #347, opened 2024-09-04
[FMV] Unify sha1, sha2.

labrinea:
According to ArmARM:

SHA2, bits [15:12]
0b0000 No SHA2 instructions implemented.
FEAT_SHA256 implements the functionality identified by the value 0b0001. If the value of ID_AA64ISAR0_EL1.SHA1 is 0b0000, this field must have the value 0b0000.

SHA1, bits [11:8]
0b0000 No SHA1 instructions implemented.
FEAT_SHA1 implements the functionality identified by the value 0b0001. If the value of ID_AA64ISAR0_EL1.SHA2 is 0b0000, this field must have the value 0b0000.

To my understanding this means you can't have one without the other.

---
name: Pull request
about: Technical issues, document format problems, bugs in scripts or feature proposal.

---

**Thank you for submitting a pull request!**

If this PR is about a bugfix:

Please use the bugfix label and make sure to go through the checklist below.

If this PR is about a proposal:

We are looking forward to evaluate your proposal, and if possible to
make it part of the Arm C Language Extension (ACLE) specifications.

We would like to encourage you reading through the [contribution
guidelines](https://github.com/ARM-software/acle/blob/main/CONTRIBUTING.md), in particular the section on [submitting
a proposal](https://github.com/ARM-software/acle/blob/main/CONTRIBUTING.md#proposals-for-new-content).

Please use the proposal label.

As for any pull request, please make sure to go through the below
checklist.

Checklist: (mark with ``X`` those which apply)

* [ ] If an issue reporting the bug exists, I have mentioned it in the
      PR (do not bother creating the issue if all you want to do is
      fixing the bug yourself).
* [ ] I have added/updated the `SPDX-FileCopyrightText` lines on top
      of any file I have edited. Format is `SPDX-FileCopyrightText:
      Copyright {year} {entity or name} <{contact informations}>`
      (Please update existing copyright lines if applicable. You can
      specify year ranges with hyphen , as in `2017-2019`, and use
      commas to separate gaps, as in `2018-2020, 2022`).
* [ ] I have updated the `Copyright` section of the sources of the
      specification I have edited (this will show up in the text
      rendered in the PDF and other output format supported). The
      format is the same described in the previous item.
* [ ] I have run the CI scripts (if applicable, as they might be
      tricky to set up on non-*nix machines). The sequence can be
      found in the [contribution
      guidelines](https://github.com/ARM-software/acle/blob/main/CONTRIBUTING.md#continuous-integration). Don't
[...]

PR #352, opened 2024-10-03
[FMV] Unify aes with pmull and sve2-aes with sve2-pmull128.

labrinea:
I originally tried splitting these features (see relevant pull request https://github.com/llvm/llvm-project/pull/110816), but the following came to my attention:

According to https://developer.arm.com/documentation/ddi0487/latest Arm Architecture Reference Manual for A-profile architecture:

D23.2.83 ID_AA64ZFR0_EL1, SVE Feature ID Register 0

ID_AA64ZFR0_EL1.AES, bits [7:4]

> FEAT_SVE_AES implements the functionality identified by the value 0b0001.
> FEAT_SVE_PMULL128 implements the functionality identified by the value 0b0010.
> The permitted values are 0b0000 and 0b0010.

Andrew Carlotti suggests that the same applies for ID_AA64ISAR0_EL1.AES (https://github.com/llvm/llvm-project/pull/110816#issuecomment-2388410290)

D19.2.61 ID_AA64ISAR0_EL1, AArch64 Instruction Set Attribute Register 0

ID_AA64ISAR0_EL1.AES, bits [7:4]

> FEAT_AES implements the functionality identified by the value 0b0001.
> FEAT_PMULL implements the functionality identified by the value 0b0010.
> From Armv8, the permitted values are 0b0000 and 0b0010.

This was removed from the latest release of the Arm Architecture Reference Manual, but it appears to be a mistake that was not intended to relax the architecture constraints. The discrepancy has been reported.

---
name: Pull request

---

If this PR is about a bugfix:

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -417,6 +417,8 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 
 * Unified Function Multi Versioning features ssbs and ssbs2.
 * Unified Function Multi Versioning features memtag and memtag2.
+* Unified Function Multi Versioning features aes and pmull.
+* Unified Function Multi Versioning features sve2-aes and sve2-pmull128.
 
 ### References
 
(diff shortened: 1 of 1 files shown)


labrinea:
@vhscampos @andrewcarlotti @tmatheson-arm

labrinea:
Ping

DanielKristofKiss:
IMHO from the user(developer) point of view the 'pmull' feature could be useful outside of `aes` so in cases it could be more expressive to say `pmull` instead of `aes` when really only pmull is used actually from the feature set. 
(it would be fine if that is just an alias to aes internally)

andrewcarlotti:
This looks good to me. I don't think there's any point creating aliases when the corresponding command line flags have existed without any need for aliases for many years.

labrinea:
This case is similar to feat_sha1 and feat_sha256 (you can't have one without having the other), which we decided to unify: https://github.com/ARM-software/acle/pull/347. I think we should do the same here.

tmatheson-arm:
Agree, now

> This case is similar to feat_sha1 and feat_sha256 (you can't have one without having the other), which we decided to unify: #347. I think we should do the same here.

Agree, now that we have set the precedent for keeping FMV aligned with the command line options, we should stick with it. If a need arises for specifying pmull without aes then they can be added to both (command line and FMV) later.

tmatheson-arm (approved):

PR #355, opened 2024-10-29
[FMV][AArch64] Remove features which expose non exploitable runtime behavior.

labrinea:
Feature `rpres` allows an increase in the precision of the single-precision
floating-point reciprocal estimate and reciprocal square root estimate
from an 8-bit mantissa to a 12-bit mantissa depending on the value FPCR.AH
of the Floating-point Control Register.

Similarly, `ebf16` allows existing floating-point instructions (BFDOT, BFMMLA,
BFMOPA, BFMOPS, and BFVDOT) to change numeric behaviour depending on the
value FPCR.EBF of the Floating-point Control Register.

Feature `memtag3` allows Tag Check Faults to change behaviour depending on
the value SCTLR_ELx.{TCF, TCF0} of the System Control Register.

The runtime detection in FMV does not examine the content of control
registers, therefore runtime dispatch cannot be based on that. One may
argue there is value in altering the control register from a version,
for example "msr dit, \#1" if the feature is available on hardware. The
registers FPCR and SCTLR_ELx can be accessed in the absence of rpres,
ebf16, and memtag3, making it hard to justify adding them to the compiler.

---
name: Pull request

---

If this PR is about a bugfix:

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -419,6 +419,7 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * Unified Function Multi Versioning features memtag and memtag2.
 * Unified Function Multi Versioning features aes and pmull.
 * Unified Function Multi Versioning features sve2-aes and sve2-pmull128.
+* Removed Function Multi Versioning features ebf16, memtag3, and rpres.
 * Fixed range of operand `o0` (too small) in AArch64 system register designations.
 * Fixed SVE2.1 quadword gather load/scatter store intrinsics.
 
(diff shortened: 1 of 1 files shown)


labrinea:
@andrewcarlotti @DanielKristofKiss @tmatheson-arm @Wilco1

andrewcarlotti:
LGTM

andrewcarlotti:
The same justification applies to dit - the behaviour is determined by a control bit, and the architecture feature merely allows this bit to be switched on.

Wilco1:
Looks good to me.

jroelofs:
I don't agree with the justification for removal of `dit`: you still need the ISA feature presence check around `msr dit, x0` instructions that toggle the feature, and prior to this change, FMV _could_ be used for that.

labrinea:
> I don't agree with the justification for removal of dit: you still need the ISA feature presence check around msr dit, x0 instructions that toggle the feature, and prior to this change, FMV could be used for that.

I kinda see your point, but the value of CPSR.DIT cannot be detected at runtime in order to choose a function version based on that. Even if it was the case I still can't think of a meaningful example. Perhaps something like the following snippet?
```cpp
__attribute__((target_version("dit"))) bool enableDIT(void) { asm volatile ("msr dit, x0"); return true; }
__attribute__((ttarget_version("default"))) bool enableDIT(void) { return false; }

if (enableDIT())
  // do something
else
  // do something else
```
I am not sure what kind of decisions can be made in the if-else logic based on this.

jroelofs:
I feel this should be sufficient justification:
```
__attribute__((target_version("dit"))) void enableDITIfExists(void) { asm volatile ("msr dit, x0"); }
__attribute__((ttarget_version("default"))) void enableDITIfExists(void) { }
```

jroelofs:
This solves a very real problem, by the way: https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms#Enable-DIT-for-constant-time-cryptographic-operations

jroelofs:
>> I don't agree with the justification for removal of dit: you still need the ISA feature presence check around msr dit, x0 instructions that toggle the feature, and prior to this change, FMV could be used for that.
>
> I kinda see your point, but the value of CPSR.DIT cannot be detected at runtime in order to choose a function version based on that

By "ISA feature presence check" I mean whether or not the cpu supports DIT at all, not runtime detection of whether or not DIT is enabled. Those two axes are almost orthogonal, but FMV is only useful for the former, so I agree that it is not useful for the latter.

vhscampos:
Thanks. Merging now

PR #363, opened 2024-11-21
Minor modification to FMV rules for scope and signatures

AlfieRichardsArm:
Hi all,

While attempting to implement FMV in the GCC front-end some questions were raised that I think are worth clarifying here.

This PR changes the rules to use the default function to determine the signature and scope of the versioned function set.

This clears up some cases such as:

```C
int fn (int c = 1);
int __attribute__((target_version("sve"))) fn (int c = 2);

int bar() { return fn(); }
```
Where there are conflicting signatures and which default should be used is not clear at the moment.

```C
int fn (int c[]);
int __attribute__((target_version("default"))) fn (int c[1]) {
}
int __attribute__((target_version("sve"))) fn (int c[2]) {
}
```
Where if this should be considered a conflicting signature is not clear.

```C
int __attribute__((target_version("default"))) fn (int x) {
    return 1;
}

void bar () {
    int __attribute__((target_version("sve2"))) fn (int);
    fn(1);
}
```
Where the scope of multi-versioned functions differs.

And
```C
// TU 1
#import TU2

int fn (int c = 1);

int bar() { return fn(); }

// TU 2

int __attribute__((target_version("sve"))) fn (int c = 2);
int __attribute__((target_version("sve2"))) fn (int c = 2);

int bar() { return fn(); }
```
Where it is possible calls in different TU's could use different default argument values.

---
name: Pull request

---

If this PR is about a bugfix:

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -433,6 +433,8 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
   than the [NEON-SVE bridge](#neon-sve-bridge) intrinsics.
 * Removed extraneous `const` from SVE2.1 store intrinsics.
 * Added [`__arm_agnostic`](#arm_agnostic) keyword attribute.
+* Refined function versioning scope and signature rules to use the default
+  version scope and signature.
 
 ### References
 
@@ -2675,10 +2677,12 @@ The following attributes trigger the multi version code generation:
 `__attribute__((target_version("name")))` and
 `__attribute__((target_clones("name",...)))`.
 
+* Functions are allowed to have the same name and signature when
+  annotated with these attributes.
 * These attributes can be mixed with each other.
+* `name` is the dependent features from the tables below.
 * The `default` version means the version of the function that would
   be generated without these attributes.
-* `name` is the dependent features from the tables below.
 * The dependent features could be joined by the `+` sign.
 * None of these attributes will enable the corresponding ACLE feature(s)
   associated to the `name` expressed in the attribute.
@@ -2687,21 +2691,46 @@ The following attributes trigger the multi version code generation:
 * If only the `default` version exist it should be linked directly.
 * FMV may be disabled in compile time by a compiler flag. In this
   case the `default` version shall be used.
+* All function versions must be declared at the same scope level.
+* The default version signature is the signature for calling
+  the multiversioned functions. Therefore, a versioned function
+  cannot be called unless the declaration of the default version
+  is visible in the scope of the call site.
+* Non-default versions shall have a type that is convertible to the
+  type of the default version.
+* All the function versions must be declared at the translation
+  unit in which the definition of the default version resides.
 
 The attribute `__attribute__((target_version("name")))` expresses the
 following:
 
-* when applied to a function it becomes one of the versions. Function
-  with the same name may exist with multiple versions in the same
-  or in different translation units.
+* When applied to a function it becomes one of the versions.
+* Multiple function versions may exist in the same or in different
+  translation units.
 * One `default` version of the function is required to be provided
   in one of the translation units.
   * Implicitly, without this attribute,
   * or explicitly providing the `default` in the attribute.
-* All instances of the versions shall share the same function
-  signature and calling convention.
-* All the function versions must be declared at the translation
-  unit in which the definition of the default version resides.
+
+For example, the below is valid and 2 is used as the default
+value for `c` when calling the multiversioned function `f`.
+
+```cpp
+int __attribute__((target_version("simd"))) f (int c = 1);
+int __attribute__((target_version("default"))) f (int c = 2);
+int __attribute__((target_version("sve"))) f (int c = 3);
+
+int g() { return f(); }
+```
+
+Additionally, the below is not valid as the two statements declare
+the same entity (the `default` version of `f`) with conflicting
+signatures.
+
+```cpp
+int f (int c = 1);
+int __attribute__((target_version("default"))) f (int c = 2);
+```
 
 The attribute `__attribute__((target_clones("name",...)))` expresses the
 following:


main/acle.md
```diff
@@ -2682,6 +2683,10 @@ The following attributes trigger the multi version code generation:
 * If only the `default` version exist it should be linked directly.
 * FMV may be disabled in compile time by a compiler flag. In this
   case the `default` version shall be used.
+* Scope for a group of multiversioned functions is the scope of the default
```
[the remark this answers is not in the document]
AlfieRichardsArm: I completely agree we want to be able to compile in absence of default declaration. I wrote this to exclude the calling or referring to the dispatched function without the default in scope (rather than the declaration).

My model for this is that the default function _is_ the function, and the extra versions amend that function with different versions for different architecture.
 
This way, if you were to remove all the other version implementations and declarations the code base would still compile.

I think a case like https://godbolt.org/z/qW7hx58bf is emblematic of this point. In this TU there is no indication there is a default version, I think this should require a default version to be declared.

Even in the case you cited (https://github.com/llvm/llvm-project/pull/84405#discussion_r1525583647) they do include the declaration of the default version in all TU's, I believe because it is the idiomatic way to do it?

main/acle.md
```diff
@@ -2696,10 +2701,12 @@ following:
   in one of the translation units.
   * Implicitly, without this attribute,
   * or explicitly providing the `default` in the attribute.
-* All instances of the versions shall share the same function
-  signature and calling convention.
```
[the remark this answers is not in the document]
AlfieRichardsArm: I agree the wording should be improved, but I believe there is a need for an adjustment here.
This was specifically something requested by upstream C maintainers in GCC.

The justification is that we likely don't want to require exactly the same signatures, as it
is common for declarations and implementations to differ in signature while still being compatible:

Something like
```C++
int __attribute__((target_version("default"))) fn (int x[]);

int __attribute__((target_version("default"))) fn (int x[5]) {
    return 1;
}
```

Or:
```C++
int __attribute__((target_version("default"))) fn (bool x = true);

int __attribute__((target_version("default"))) fn (bool x) {
    return 1;
}
```
For a more CPP example.

(LLVM is happy with both these currently)

But if this is allowed at some level then we need to define if information for one version should be shared with others, eg a case like:

```C++
int __attribute__((target_version("default"))) fn (bool c = true);
int __attribute__((target_version("sve"))) fn (bool c);

int __attribute__((target_version("default"))) fn (bool c) {
    return 1;
}

int bar() {
    fn();
}
```
(Again allowed by LLVM)
Should this default to true for the SVE version? I would assume so but its not stipulated?

We could disallow it completely, but that seems overly restrictive and counter to C and C++ intuitions.

main/acle.md
```diff
@@ -2682,6 +2683,10 @@ The following attributes trigger the multi version code generation:
 * If only the `default` version exist it should be linked directly.
 * FMV may be disabled in compile time by a compiler flag. In this
   case the `default` version shall be used.
+* Scope for a group of multiversioned functions is the scope of the default
+  version declarations.
+* Dispatiching routines can only be optimized out within the translation unit
```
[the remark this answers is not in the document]
AlfieRichardsArm: That makes sense. I'll remove this.

main/acle.md
```diff
@@ -2696,10 +2701,12 @@ following:
...
   * Implicitly, without this attribute,
   * or explicitly providing the `default` in the attribute.
-* All instances of the versions shall share the same function
-  signature and calling convention.
+* All instances of the versions shall share the same calling convention 
+* All function declarations shall share compatible function signatures 
+  that are combined to a single signature for all versions.
 * All the function versions must be declared at the translation
   unit in which the definition of the default version resides.
+* All function declarations must be at file scope level.
```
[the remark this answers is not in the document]
AlfieRichardsArm: Hmm namespaces are a good point (I was looking at this from the C point of view)

I mostly wanted to exclude the case with the versioned function declaration in a function body.
I think it creates the impression that a call in the function will be able to access different versions that outside it?

Like in the example you gave, the resolver calls the `sve2` dispatched version of the function, but it isn't visible to the default version so surely it shouldn't? But what should it do?

main/acle.md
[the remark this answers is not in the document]
AlfGalf: Oh wow, okay. That example gives me really bad mojo, and I would advocate we should disallow that if possible, though I wont die on that hill. 
I can live with the default signature dominating, but request we document that in the ACLE.

I would advocate for documenting it in the ACLE. I think the purpose of the ACLE is to document this sort of edge case so we can ensure matched behavior between compilers, and this point came up because my natural inclination when implementing it in GCC was to go another direction. 

This was specifically requested by the GCC C front end maintainer for what I imagine was very much this sort of reason.

main/acle.md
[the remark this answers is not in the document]
AlfGalf: I acknowledge that I'm very junior in my career and that I am may be missing significant nuance in this.

I see that argument, but I think I disagree. I think we do need to change something. 

If a multi versioned functions signature is is dominated by the default (ie, that's the canonical signature) than I dont think it makes sense to allow calling that function without that signature in scope.

For example
```C
// TU 1
#import TU2

int fn (int c = 1);

int bar() { return fn(); }

// TU 2

int __attribute__((target_version("sve"))) fn (int c = 2);
int __attribute__((target_version("sve2"))) fn (int c = 2);

int bar() { return fn(); }
```
Would be valid and the calls would have different defaults.

I think the above is quite a good example of why the default declaration determining scope is a good idea.

I also don't think its unreasonable to require the other TU's to have hoisted/shared default declarations. That is the standard for C/C++ functions called in multiple files, and if foo calls bar, and bar calls foo, this would also be required no? It seems idiomatic to require it?

main/acle.md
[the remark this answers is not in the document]
AlfGalf: Agreed, I think that works :)
I will update the PR

main/acle.md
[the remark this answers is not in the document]
jroelofs: I think I would be okay with a rule that only allows calling a multi-versioned function if you can see the default declaration: we can add diagnostics + fixits that help users do the right thing.

AlfieRichardsArm:
I have updated this PR with hopefully better wording more inline with @labrinea 's suggestions.

main/acle.md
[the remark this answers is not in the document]
AlfieRichardsArm: Ah okay I see, thank you. I have changed the patch to document this rule a little clearer.

As this came up with the GCC front end maintainers I would prefer to clarify this case.

main/acle.md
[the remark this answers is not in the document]
AlfieRichardsArm: Thank you for the suggestion, I have updated the PR wording to be closer to that.

main/acle.md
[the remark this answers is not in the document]
labrinea: The wording seems better. Perhaps it is worth adding the example https://godbolt.org/z/Enn15Y93e without the codegen but explaining that the value 2 will be passed when calling.

main/acle.md
```diff
@@ -2699,8 +2702,10 @@ following:
   in one of the translation units.
   * Implicitly, without this attribute,
   * or explicitly providing the `default` in the attribute.
-* All instances of the versions shall share the same function
-  signature and calling convention.
+* The default version signature is the definitive signature for calling
+  the multiversioned functions.
+  * All other instances of the versions shall share the same calling
```
labrinea, line 2707: The indentation seems off.

labrinea (approved):

main/acle.md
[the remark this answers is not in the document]
AlfieRichardsArm: Just to check, do you mean adding this directly to the ACLE?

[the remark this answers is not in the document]
AlfieRichardsArm: Added some now :)

main/acle.md
```diff
@@ -2699,11 +2704,41 @@ following:
...
+This case is valid as it is possible to use the
+non-default function as the value for a pointer with the type of
+the default function, as shown below.
+
+```C
+void f_foo(void);
+void (*f)(void) = f_foo;
+```
+
+Additionally, the below example is valid and 2 is passed as the
```
labrinea, line 2731: I think we can merge these two examples into one if you still want to keep the first? There is no versioned `f_foo` in your first example.
```C++
__attribute__((target_version("default"))) void f(int c = 1);
__attribute__((target_version("sve"))) void f(int);

void (*ptr)(int) = f;

void caller() {
  f(); // passes 1
  ptr(2); // passes 2
}
```

main/acle.md
```diff
@@ -2699,11 +2704,41 @@ following:
...
 
+For example, in the case below:
+```C
+__attribute__((target_version("default"))) void fn(void);
+__attribute__((target_version("foo"))) void fn(void); // f_foo below
+````
+
+This case is valid as it is possible to use the
+non-default function as the value for a pointer with the type of
+the default function, as shown below.
```
labrinea, line 2724: Perhaps you are mistaken here? We cannot explicitly refer to a specific version. Taking the address of a versioned function and assigning it into a pointer will take the address of the ifunc symbol.

main/acle.md
```diff
@@ -2699,11 +2704,41 @@ following:
   in one of the translation units.
   * Implicitly, without this attribute,
   * or explicitly providing the `default` in the attribute.
-* All instances of the versions shall share the same function
-  signature and calling convention.
+* The default version signature is the signature for calling
+  the multiversioned functions.
+* Non-default versions shall share the same calling convention
+  of the default version and have a type that is convertable to the type
```
labrinea, line 2710: spelling 'convertible'?

main/acle.md
jroelofs: Agreed. That said, I'd like to have syntax for that: https://github.com/llvm/llvm-project/issues/84094

main/acle.md
```diff
@@ -2699,11 +2704,41 @@ following:
...
+  of the default version.
+
 * All the function versions must be declared at the translation
   unit in which the definition of the default version resides.
 
+For example, in the case below:
+```C
+__attribute__((target_version("default"))) void fn(void);
+__attribute__((target_version("foo"))) void fn(void); // f_foo below
```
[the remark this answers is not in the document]
AlfieRichardsArm: Yup, sounds good.

main/acle.md
AlfieRichardsArm: This example is intended to clarify what is meant by `Non-default versions shall [..] have a type that is convertible to the type of the default version`. 
I intended to show that a non-default function has "convertible" signature if a function of the non-default signature could be assigned to a function pointer with the default functions signature.

This was a suggestion by Richard Sandiford, that we should use existing function pointer rules to define signature/type compatibility. However, I struggled to formalize it and clearly this example has is not clear for you two, let alone non-experts, so unless you have a way to write this clearer, I will remove it.

main/acle.md
AlfieRichardsArm: Thank you!

main/acle.md
AlfieRichardsArm: I think I'll remove the first one. Thank you!

main/acle.md
```diff
@@ -2699,11 +2704,26 @@ following:
...
 
+For example, the below is valid and 2 is passed as the
+value for `c` when calling `fn`.
+
+```C
+int __attribute__((target_version("simd"))) fn (int c = 1);
+int __attribute__((target_version("default"))) fn (int c = 2);
+int __attribute__((target_version("sve"))) fn (int c = 3);
+
+int baz() { return fn(); }
```
labrinea, line 2724: Let's rename baz to something else. I suggest either use naming scheme `g` for caller and `f` for callee. Or just rename baz to `caller`. Otherwise LGTM

labrinea (approved):

main/acle.md
AlfieRichardsArm: Awesome, done! Thank you so much for your help with this.

vhscampos:
If no objections are raised, let's merge this on the 23 December.

main/acle.md
```diff
@@ -2699,11 +2704,26 @@ following:
...
+For example, the below is valid and 2 is passed as the
+value for `c` when calling `f`.
+
+```C
+int __attribute__((target_version("simd"))) f (int c = 1);
+int __attribute__((target_version("default"))) f (int c = 2);
+int __attribute__((target_version("sve"))) f (int c = 3);
+
+int g() { return f(); }
+```
```
rsandifo-arm, line 2725: Just a suggestion, but it might also be worth adding an example to say that:
```
int f (int c = 1);
int __attribute__((target_version("default"))) f (int c = 2);
```
is not valid, because the two statements are taken to be declaring the same entity.

jroelofs (approved):

main/acle.md
AlfieRichardsArm: Thank you for the suggestions, Ive added this.

main/acle.md
```diff
@@ -2699,11 +2704,34 @@ following:
...
+  of the default version and have a type that is convertible to the type
+  of the default version.
+
 * All the function versions must be declared at the translation
   unit in which the definition of the default version resides.
 
+For example, the below is valid and 2 is passed as the
+value for `c` when calling `f`.
+
+```c
```
labrinea, line 2719: NIT: Not sure if it matters, but I think default initialization of arguments is only valid in C++ code, so perhaps the snipped should be marked accordingly?

AlfieRichardsArm: 👍 Done

labrinea: I think the markdown annotation for C++ is cpp. You can use build_with_docker.sh to generate locally a pdf version of the document. This way you can double check what the right syntax is.

AlfieRichardsArm: Ah apologies. I have changed this to use `cpp`, checked `build_with_docker.sh`, and rebased onto main.

main/acle.md
```diff
@@ -2698,11 +2703,34 @@ following:
   in one of the translation units.
   * Implicitly, without this attribute,
   * or explicitly providing the `default` in the attribute.
-* All instances of the versions shall share the same function
-  signature and calling convention.
+* The default version signature is the signature for calling
+  the multiversioned functions.
+* Non-default versions shall share the same calling convention
+  of the default version and have a type that is convertible to the type
```
labrinea, line 2709: "the same calling convention of the default version"

s/of/as

main/acle.md
```diff
@@ -2687,6 +2689,9 @@ The following attributes trigger the multi version code generation:
 * If only the `default` version exist it should be linked directly.
 * FMV may be disabled in compile time by a compiler flag. In this
   case the `default` version shall be used.
+* The scope for calling the versioned function is the scope of the
+  default version.
```
labrinea, line 2693: Perhaps rephrase to make it easier for the reader? "A versioned function cannot be called unless the declaration of the default version is visible in the scope of the call site".

AlfieRichardsArm:
@labrinea Thank you for the feedback, I addressed them and in doing so moved some stuff around slightly so I hope it makes more sense now.

main/acle.md
```diff
@@ -2687,6 +2689,15 @@ The following attributes trigger the multi version code generation:
 * If only the `default` version exist it should be linked directly.
 * FMV may be disabled in compile time by a compiler flag. In this
   case the `default` version shall be used.
+* A versioned function cannot be called unless the declaration of
+  the default version in visible in the scope of the call site.
```
labrinea, line 2693: typo `is` visible

labrinea:
May we move the sentence https://github.com/ARM-software/acle/blob/main/main/acle.md?plain=1#L2695 here https://github.com/ARM-software/acle/blob/main/main/acle.md?plain=1#L2677 and rephase it as "Functions are allowed to have the same name when annotated with these attributes." Then in https://github.com/ARM-software/acle/blob/main/main/acle.md?plain=1#L2695 we could say "when applied to a function it becomes one of the versions. Multiple function versions may exist in the same or in different translation units."

Now regarding the order of bullet points I think it could be improved to help the reader. For example the second bullet point refers to the default version a bit suddenly. The third bullet point refers to the table of features which mentions 'default' for the first time in this document.

Also some of thebullet points refer to FMV semantics whereas others specifically refer to the FMV attributes. Perhaps it's worth separating those.

This can be done in a separate commit.

AlfieRichardsArm:
I agree the bullets can be made clearer. I will look into making another PR for that work shortly.

I addressed your feedback and while reading through made a few other touches to parts hopefully within scope for this change to make it a little clearer.

main/acle.md
```diff
@@ -2687,21 +2691,46 @@ The following attributes trigger the multi version code generation:
 * If only the `default` version exist it should be linked directly.
 * FMV may be disabled in compile time by a compiler flag. In this
   case the `default` version shall be used.
+* A versioned function cannot be called unless the declaration of
+  the default version is visible in the scope of the call site.
+* All function versions must be declared at the same scope level.
+* The default version signature is the signature for calling
+  the multiversioned functions.
```
labrinea, line 2698: Under the same bullet: "Therefore a versioned function cannot be called unless the declaration of
  the default version is visible in the scope of the call site." Now we can remove the previous bullet. The reasoning makes more sense if these two points are glued together.

AlfieRichardsArm: That makes sense! Done

labrinea (approved):

main/acle.md
```diff
@@ -2675,10 +2677,12 @@ The following attributes trigger the multi version code generation:
 `__attribute__((target_version("name")))` and
 `__attribute__((target_clones("name",...)))`.
 
+* Functions are allowed to have the same name and signature when
+  annotated with these attributes.
 * These attributes can be mixed with each other.
+* `name` is the dependent features from the tables below.
```
philgee-oss, line 2683: Don't use below or above, because they can be ambiguous. Best to cross-reference to the section containing the tables (Mapping?)

main/acle.md
```diff
@@ -2675,10 +2677,12 @@ The following attributes trigger the multi version code generation:
...
 * The `default` version means the version of the function that would
   be generated without these attributes.
-* `name` is the dependent features from the tables below.
 * The dependent features could be joined by the `+` sign.
 * None of these attributes will enable the corresponding ACLE feature(s)
```
philgee-oss, line 2687: attributes enable

main/acle.md
```diff
@@ -2687,21 +2691,46 @@ The following attributes trigger the multi version code generation:
 * If only the `default` version exist it should be linked directly.
 * FMV may be disabled in compile time by a compiler flag. In this
```
philgee-oss, line 2692: FMV might

main/acle.md
```diff
@@ -2687,21 +2691,46 @@ The following attributes trigger the multi version code generation:
...
   case the `default` version shall be used.
```
philgee-oss, line 2693: case, the

main/acle.md
```diff
@@ -2687,21 +2691,46 @@ The following attributes trigger the multi version code generation:
...
 * One `default` version of the function is required to be provided
   in one of the translation units.
   * Implicitly, without this attribute,
   * or explicitly providing the `default` in the attribute.
-* All instances of the versions shall share the same function
-  signature and calling convention.
-* All the function versions must be declared at the translation
-  unit in which the definition of the default version resides.
+
+For example, the below is valid and 2 is used as the default
```
philgee-oss, line 2715: The following example is valid

main/acle.md
```diff
@@ -2687,21 +2691,46 @@ The following attributes trigger the multi version code generation:
...
+
+```cpp
+int __attribute__((target_version("simd"))) f (int c = 1);
+int __attribute__((target_version("default"))) f (int c = 2);
+int __attribute__((target_version("sve"))) f (int c = 3);
+
+int g() { return f(); }
+```
+
+Additionally, the below is not valid as the two statements declare
```
philgee-oss, line 2726: Additionally, the following example is not valid because

AlfieRichardsArm:
@philgee-oss Hi, thanks for the feedback, however this has been merged as of 6th Jan.
I can open a new PR and address these if you think its worth it?

vhscampos:
> @philgee-oss Hi, thanks for the feedback, however this has been merged as of 6th Jan. I can open a new PR and address these if you think its worth it?

Hi @AlfieRichardsArm , don't worry, I will fix all the issues raised by Phil

PR #369, opened 2024-12-09
Add svdot[_n_f16_mf8]_fpm and svdot[_n_f32_mf8]_fpm

momchil-velikov:
Accidentally omitted.

---
name: Pull request

---

If this PR is about a bugfix:

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.


andrewcarlotti:
I don't think we want these - the other `svdot` intrinsics don't have an `_n` variant either, and I don't think it makes sense to add one since the indexed segment for these instructions consists of either two or four elements, unlike the single indexed element used in `MLA` instructions.

momchil-velikov:
> I don't think we want these - the other `svdot` intrinsics don't have an `_n` variant either, and I don't think it makes sense to add one since the indexed segment for these instructions consists of either two or four elements, unlike the single indexed element used in `MLA` instructions.

I don't see the relevance of the indexed forms of the instructions.
There are seven other  `*dot_n` intrinsics and also other `*_n` intrinsics.

rsandifo-arm:
LGTM, but could you also change:

````
#### UDOT, SDOT, FDOT (vectors)

Multi-vector dot-product (2-way)

``` c
  // Variants are also available for _s32_s16 and _u32_u16
  svfloat32_t svdot[_f32_f16](svfloat32_t zda, svfloat16_t zn,
                              svfloat16_t zm);
  ```
````

while you're there?

momchil-velikov:
> LGTM, but could you also change:
> 
> ```
> #### UDOT, SDOT, FDOT (vectors)
> 
> Multi-vector dot-product (2-way)
> 
> ``` c
>   // Variants are also available for _s32_s16 and _u32_u16
>   svfloat32_t svdot[_f32_f16](svfloat32_t zda, svfloat16_t zn,
>                               svfloat16_t zm);
>   ```
> ```
> 
> while you're there?

What needs to change?

andrewcarlotti:
Ah - so https://developer.arm.com/architectures/instruction-sets/intrinsics/ says that there are dot_n intrinsics for SVE but not for Neon, but they aren't listed in acle.md. I just checked that they are also defined in GCC.

I think Richard is asking you to add the existing indexed versions to the list of SVE intrinsics in acle.md.

Given that we actually do have existing dot_n intrinsics for SVE, I'm fine with adding these ones as well. But I still might have preferred to have none of them. 
For comparison, the main justification (in my opinion) for including mla_n intrinsics is that they correspond to a single architecture instruction (the indexed for of the instruction with index 0). This justification doesn't apply to the dot_n intrinsics, where we still have to duplicate the scalar input into a vector.

rsandifo-arm:
> Given that we actually do have existing dot_n intrinsics for SVE, I'm fine with adding these ones as well. But I still might have preferred to have none of them.
> For comparison, the main justification (in my opinion) for including mla_n intrinsics is that they correspond to a single architecture instruction (the indexed for of the instruction with index 0). This justification doesn't apply to the dot_n intrinsics, where we still have to duplicate the scalar input into a vector.

That wasn't the reason for adding the extra `_n` intrinsics to SVE.  Instead it was that:
* SVE has (relatively) more immediate forms than Advanced SIMD, and we wanted those immediate forms to be available without an explicit dup.
* The immediates are somewhat irregular (e.g. 0 or 1 for fmax, 0.5 or 2 for fmul, 0.5 or 1 for fadd), so we wanted to hide the exact ranges from the user.  We therefore allowed the scalar arguments to be any immediate, not just in-range immediates, with the compiler synthesising a constant vector where necessary.
* That in turn meant that it was more natural not to require an immediate at all, but instead allow any scalar.
* Then it seemed natural to provide these immediate-like scalars in all contexts where an immediate would make conceptual sense.  That gives a more regular interface, and avoids users having to remember which instructions have immediate forms and which don't.

andrewcarlotti:
> That wasn't the reason for adding the extra `_n` intrinsics to SVE. Instead it was that:
> 
> * SVE has (relatively) more immediate forms than Advanced SIMD, and we wanted those immediate forms to be available without an explicit dup.
> * The immediates are somewhat irregular (e.g. 0 or 1 for fmax, 0.5 or 2 for fmul, 0.5 or 1 for fadd), so we wanted to hide the exact ranges from the user.  We therefore allowed the scalar arguments to be any immediate, not just in-range immediates, with the compiler synthesising a constant vector where necessary.
> * That in turn meant that it was more natural not to require an immediate at all, but instead allow any scalar.
> * Then it seemed natural to provide these immediate-like scalars in all contexts where an immediate would make conceptual sense.  That gives a more regular interface, and avoids users having to remember which instructions have immediate forms and which don't.

Thanks for the context - it's interesting that the same conclusions arise from considering instructions with immediates instead of instructions with index zero.

CarolineConcatto:
Hey Momchil,

momchil-velikov:
> Hey Momchil, I believe you forgot to address Richard's comment: LGTM, but could you also change:

I most certianly have not forgotten to address Richard's comment.

momchil-velikov:
Also I don't see why this patch should be blocked by any other unrelated change.

rsandifo-arm:
OK, if you think that the other instance of an `svdot` with a missing `_n` form should be a separate PR, I'll create one for that.

CarolineConcatto:
It looks like Richard also made the change:
https://github.com/ARM-software/acle/pull/378/files
So I can approve yours and his.

CarolineConcatto (approved):

Issue #403, opened 2025-06-23
[proposal] Add syntax for expressing priority of FMV target expressions


AlfieRichardsArm:
## Summary

This proposal adds support for explicitly expressing the priority of function versions, in the target strings.

## Example use case

For the case below, the default ordering has the `sve` feature as higher priority then `dotprod`, meaning for a target with both `sve` and `dotprod` the `sve` version would be selected. However, it may be that the dotprod version should be priority in this case.

```c
typedef struct {
  int x;
  int y;
  int z;
  int w;
} Vec4;

[[target_clone("default", "sve")]]
Vec4 dotproduct (Vec4 a, Vec4 b)
{
  // do a dot product
}

[[target_version("dotprod")]]
Vec4 dotproduct (Vec4 a, Vec4 b)
{
  // Use dotprod intrinsics
}
```

## Motivation

This enhancement was a natural result of considering the version priority rules. In specifying the default ordering, cases such as above, arose where the default priority rules would be a poor choice.

## Possible solutions

There are several ways to solve this. 

#### 1. Add dummy features "priorityA", "priorityB", ..., 

These features would have no effect on the versioned function, other than to change how they are ordered.
These versions would be higher priority than any other feature, so would override any default ordering.

Then the above would be:

```c
typedef struct {
  int x;
  int y;
  int z;
  int w;
} Vec4;

[[target_clone("default", "sve")]]
Vec4 dotproduct (Vec4 a, Vec4 b)
{
  // do a dot product
}

[[target_version("dotprod+priorityA")]]
Vec4 dotproduct (Vec4 a, Vec4 b)
{
  // Use dotprod intrinsics
}
```

Could also use "priority1", "priority2", ...

This is similar to what was done for other targets (https://github.com/riscv-non-isa/riscv-c-api-doc/pull/85/files).

#### 2. Label all the versions of a function

Another option is too support explicitly stating the order of all versions.

Something like:
```c
typedef struct {
  int x;
  int y;
  int z;
  int w;
} Vec4;

[[target_clone("P3:default", "P2:sve")]]
Vec4 dotproduct (Vec4 a, Vec4 b)
{
  // do a dot product
}

[[target_version("P1:dotprod+priorityA")]]
Vec4 dotproduct (Vec4 a, Vec4 b)
{
  // Use dotprod intrinsics
}
```

This seems like it introduces many more edge cases and complexity over suggestion 1 with little gain other than explicitness.

AlfieRichardsArm:
[...]

PR #371, opened 2024-12-10
Add FMV feature names priority[ABCDE]

alicecarlotti:
These feature names allow developers to specify a priority order for their
function versions that differs from the default priority order for the
specified architectural features.

This builds upon #370.

---
name: Pull request
about: Technical issues, document format problems, bugs in scripts or feature proposal.

---

**Thank you for submitting a pull request!**

If this PR is about a bugfix:

Please use the bugfix label and make sure to go through the checklist below.

If this PR is about a proposal:

We are looking forward to evaluate your proposal, and if possible to
make it part of the Arm C Language Extension (ACLE) specifications.

We would like to encourage you reading through the [contribution
guidelines](https://github.com/ARM-software/acle/blob/main/CONTRIBUTING.md), in particular the section on [submitting
a proposal](https://github.com/ARM-software/acle/blob/main/CONTRIBUTING.md#proposals-for-new-content).

Please use the proposal label.

As for any pull request, please make sure to go through the below
checklist.

Checklist: (mark with ``X`` those which apply)

* [ ] If an issue reporting the bug exists, I have mentioned it in the
      PR (do not bother creating the issue if all you want to do is
      fixing the bug yourself).
* [ ] I have added/updated the `SPDX-FileCopyrightText` lines on top
      of any file I have edited. Format is `SPDX-FileCopyrightText:
      Copyright {year} {entity or name} <{contact informations}>`
      (Please update existing copyright lines if applicable. You can
      specify year ranges with hyphen , as in `2017-2019`, and use
      commas to separate gaps, as in `2018-2020, 2022`).
* [ ] I have updated the `Copyright` section of the sources of the
      specification I have edited (this will show up in the text
      rendered in the PDF and other output format supported). The
      format is the same described in the previous item.
* [ ] I have run the CI scripts (if applicable, as they might be
      tricky to set up on non-*nix machines). The sequence can be
      found in the [contribution
      guidelines](https://github.com/ARM-software/acle/blob/main/CONTRIBUTING.md#continuous-integration). Don't
      worry if you cannot run these scripts on your machine, your
      patch will be automatically checked in the Actions of the pull
      request.
* [ ] I have added an item that describes the changes I have
      introduced in this PR in the section **Changes for next
      release** of the section **Change Control**/**Document history**
      of the document. Create **Changes for next release** if it does
      not exist. Notice that changes that are not modifying the
      content and rendering of the specifications (both HTML and PDF)
      do not need to be listed.
[...]

PR #404, opened 2025-07-15
Add FMV priority syntax

AlfieRichardsArm:
Hi all,

This is a PR to implement option 3 of #403 proposed by @labrinea .

It adds the proposed syntax, with higher priority values taking precedence.

Happy to modify wording if needed.

---

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -2843,14 +2843,14 @@ For example, it can be implemented as:
 
 ### Target version strings
 
-A target version string is of the following form:
+A target version string has the following form:
 
 ```
 <target version string>    := 'default'
                             | <version string>
-<version string>           := <priority string> ';' <arch strings>
+<version string>           :=  <arch strings> ';' <priority string>
                             | <arch strings>
-<priority string>          := 'priority=[1-31]'
+<priority string>          := 'priority=[1-255]'
 <arch strings>             := <arch strings> '+' arch extension
                             | arch extension
 ```
@@ -2863,8 +2863,8 @@ Valid string examples are given below
 default
 dotprod
 dotprod+flagm
-priority=5;sve
-priority=23;sve2+sme2
+sve;priority=5
+sve2+sme2;priority=23
 ```
 
 ### Name mangling
@@ -2883,7 +2883,7 @@ the [[cxxabi]](#cxxabi), and it is defined as follows:
 <vendor specific suffix> := `_` followed by token obtained from the tables below and prefixed with `M`
 ```
 
-Note, priority values do not affect mangling.
+Priority values do not affect mangling.
 
 If multiple features are requested then those shall be appended in lexicographic
 order and prefixed with `M`. The mangled name shall contain a unique set of


main/acle.md
```diff
@@ -2841,6 +2841,32 @@ For example, it can be implemented as:
...
 
+### Target version strings
+
+A target version string is of the following form:
+
+```
+<target version string>    := 'default'
+                            | <version string>
+<version string>           := <priority string> ';' <arch strings>
+                            | <arch strings>
```
labrinea, line 2852: perhaps we should allow commutative syntax?
```
| <arch strings> ';' <priority string>
```

AlfieRichardsArm: Thanks for the review.

I had discussed this with @chrisbazley and Alice Carlotti (@andrewcarlotti) offline and am leaning against allowing it.

The reasoning is, I imagine this being extended later with some other `opt=value;...` options, and that the unlabeled arch string will always be a special case. For this reason I would prefer the arch string to always go at the end or the beginning. I think this makes it clearer to understand, and avoids anything like `priority=1;sve;opt=value`, which I find a bit confusing.

main/acle.md
```diff
@@ -2990,11 +3018,19 @@ the selection algorithm is platform dependent, where with platform means
...
-the versions, and selecting that version.
-4. The selection algorithm must select the highest priority versions whose
-dependent features are all available.
+3. The selection algorithm must select the the highest precedence
+version whose dependent features are all available.
+
+The precedence for two target version strings is determined by:
+
+1. If both target version strings specify a `priority`, with different values,
+then the version with the higher `priority` has precedence.
```
labrinea, line 3027: @jroelofs if I read your comment https://github.com/ARM-software/acle/pull/371#issuecomment-2573429198 correctly you'd also prefer `higher number = higher priority`?

main/acle.md
```diff
@@ -2841,6 +2841,32 @@ For example, it can be implemented as:
...
+### Target version strings
+
+A target version string is of the following form:
+
+```
+<target version string>    := 'default'
+                            | <version string>
+<version string>           := <priority string> ';' <arch strings>
+                            | <arch strings>
+<priority string>          := 'priority=[1-31]'
```
labrinea: @jroelofs with https://github.com/llvm/llvm-project/pull/150079 we have plenty of room to go beyond 5 bits of priority. Any preference?

labrinea: I personally find 6 bits the maximum sensible, but I am also happy with 5 or less.

main/acle.md
```diff
@@ -2990,11 +3018,19 @@ the selection algorithm is platform dependent, where with platform means
...
+3. The selection algorithm must select the the highest precedence
+version whose dependent features are all available.
+
+The precedence for two target version strings is determined by:
+
+1. If both target version strings specify a `priority`, with different values,
+then the version with the higher `priority` has precedence.
+2. If only one target version strings specify a `priority`, this version has
+precedence.
+3. Otherwise, if neither target version strings specify a `priority` or specify
```
chrisbazley: s/strings specify/string specifies/

and ... or both specify the same value ..

main/acle.md
```diff
@@ -2990,11 +3018,19 @@ the selection algorithm is platform dependent, where with platform means
...
-4. The selection algorithm must select the highest priority versions whose
-dependent features are all available.
+3. The selection algorithm must select the the highest precedence
+version whose dependent features are all available.
+
+The precedence for two target version strings is determined by:
+
+1. If both target version strings specify a `priority`, with different values,
+then the version with the higher `priority` has precedence.
+2. If only one target version strings specify a `priority`, this version has
```
chrisbazley: s/strings specify/string specifies/

main/acle.md
```diff
@@ -2990,11 +3018,19 @@ the selection algorithm is platform dependent, where with platform means
 CPU/Vendor/OS as in the target triplet.
 2. The selection is permanent for the
 lifetime of the process.
-3. Among any two versions, the higher priority version is determined by
-identifying the highest priority feature that is specified in exactly one of
-the versions, and selecting that version.
-4. The selection algorithm must select the highest priority versions whose
-dependent features are all available.
+3. The selection algorithm must select the the highest precedence
```
chrisbazley: 'the the'
I would rephrase this as 'select the version with the highest precedence whose'

AlfieRichardsArm: Thanks! Pushed a change resolving these!

main/acle.md
```diff
@@ -2841,6 +2841,32 @@ For example, it can be implemented as:
...
+where `arch extension` is any Names value from the Mapping table below.
+
+Valid string examples are given below
+
+```
+default
+dotprod
+dotprod+flagm
+priority=5;sve
+priority=23;sve2+sme2
```
chrisbazley: I think it might be easier to parse these strings if the 'arch strings' always came first, because otherwise it is not obvious which is the last semicolon without reading ahead. (I made similar arguments against one version of GCC's forward parameter declaration syntax in https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3433.pdf )
I also think that putting priority and other less-important configuration values ahead of the 'arch strings' is a bit unnatural to read.

AlfieRichardsArm: That is convincing to me. I will give a bit of time for anyone else to voice an opinion, but if no one objects I will reverse these and put arch string first, and other parameters after.

main/acle.md
```diff
@@ -2857,6 +2883,8 @@ the [[cxxabi]](#cxxabi), and it is defined as follows:
 <vendor specific suffix> := `_` followed by token obtained from the tables below and prefixed with `M`
 ```
 
+Note, priority values do not affect mangling.
```
chrisbazley: 'Note, ' in documentation is usually redundant.

main/acle.md
```diff
@@ -2841,6 +2841,32 @@ For example, it can be implemented as:
 #define __FUNCTION_MULTI_VERSIONING_SUPPORT_LEVEL __ARM_ACLE_VERSION(2024, 3, 0)
 ```
 
+### Target version strings
+
+A target version string is of the following form:
```
chrisbazley: "has" might be better than "is of"

main/acle.md
labrinea: ping

main/acle.md
labrinea: ping

AlfieRichardsArm:
Addressed Chris' comments. Unless there's any more discussion I'd like to move towards merging this soon?

main/acle.md
```diff
@@ -446,6 +446,7 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * Added `svdot[_n_f16_mf8]_fpm` and `svdot[_n_f32_mf8]_fpm`.
 * Added Guarded Control Stack (GCS) at
   [**Beta**](#current-status-and-anticipated-changes) quality level.
+* Add Function Multi Versioning feature priority syntax.
```
sallyarmneale, line 449: * Add Function Multi Versioning feature priority syntax.
Change to **Added**

vhscampos: Thanks Sally. I'm going to fix this.

PR #406, opened 2025-08-01
Add FEAT_PCDPHINT Support

ricbal02:
---
Hi all, 

This is a PR to propose the addition of new intrinsics required as part of FEAT_PCDPHINT.

I'm very open to suggestions on potential improvements and certainly not overly committed to the current name.
This will need approval from LLVM as well. 

Quality Level: ALP

---

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.


main/acle.md
```diff
@@ -464,6 +464,7 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 
 * Added feature test macro for FEAT_SSVE_FEXPA.
 * Added feature test macro for FEAT_CSSC.
+* Added support for Producer-consumer data placement hints.
```
jthackray: nit: perhaps either initial caps or lowercase? i.e. "producer-consumer data placement" or "Producer-Consumer Data Placement". Currently it looks a bit wonky with 1 initial cap. What does the spec use?

main/acle.md
```diff
@@ -4746,6 +4765,56 @@ stored to memory is modified by replacing the low 32 bits of
...
+make use of the `STSHH` hint instruction immediately followed by the
+associated store instruction.
+This intrinsic is available when `__ARM_FEATURE_PCDPHINT` is defined.
+
+To access these intrinsics, `<arm_acle.h>` should be included.
+
+``` c
+  void __arm_stshh_u8(void const volatile *addr,
+                      uint8_t data,
+                      unsigned int mem, /* Memory order*/
```
jthackray: super nit: extra space required before end of comment `*/`

main/acle.md
```diff
@@ -4746,6 +4765,56 @@ stored to memory is modified by replacing the low 32 bits of
...
+
+| **Memory Order**  | **Value** | **Summary**                                                                                        |
+| ------------------| --------- | -------------------------------------------------------------------------------------------------- |
+| Relaxed              | 0         | No constraints imposed on other reads or writes, only this operation's atomicity is guaranteed. |
+| Release              | 1         | No reads or writes in the current thread can be reordered after this store.                     |
+
+| **Retention Policy** | **Value** | **Summary**                                                                      |
+| -------------------- | --------- | -------------------------------------------------------------------------------- |
+| KEEP                 | 0         | Signals to retain the updated location in the local cache of the updating PE.    |
+| STRM                 | 1         | Signals to not retain the updated locationin the local cache of the updating PE. |
```
jthackray: nit: s/locationin/location in/

main/acle.md
ricbal02: Spec goes goes for lower case. So I'll go for that

main/acle.md
```diff
@@ -3574,6 +3583,16 @@ as in `__pldx`.
 `__pldx` and `__plix` arguments cache level and retention policy
 are ignored on unsupported targets.
 
+### Intent to read prefetch
+
+``` c
+  void __ir(void const volatile *addr);
```
Wilco1: How about naming it __pldir since it's a variant of data prefetch?

Wilco1 (commented):
x

main/acle.md
ricbal02: I like it, my only concern would be around lack of consistency with the spec around the instruction.(https://developer.arm.com/documentation/ddi0602/2025-06/Base-Instructions/PRFM--immediate---Prefetch-memory--immediate--?lang=en)
Don't know how much that matters, if we prefer the pldir name for ease of understanding in the compiler?

main/acle.md
```diff
@@ -4746,6 +4765,56 @@ stored to memory is modified by replacing the low 32 bits of
...
+
+These intrinsics provide an atomic store intrinsic, which will
+make use of the `STSHH` hint instruction immediately followed by the
+associated store instruction.
+This intrinsic is available when `__ARM_FEATURE_PCDPHINT` is defined.
+
+To access these intrinsics, `<arm_acle.h>` should be included.
+
+``` c
+  void __arm_stshh_u8(void const volatile *addr,
```
WonderfulVoid: The target of a store shouldn't be `const` qualified? And why is it `volatile` qualified?
And why not use the proper data type for the target pointer?
GCC atomic builtins use plain `int` for the memorder parameter. See https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html
`void __arm_stshh_u8(uint8_t *addr, uint8_t data, int memorder, int retpol);`

main/acle.md
```diff
@@ -4746,6 +4765,56 @@ stored to memory is modified by replacing the low 32 bits of
...
+                       unsigned int mem, /* Memory order */
+                       unsigned int ret); /* Retention Policy */
+```
+The first argument in these intrinsics is a pointer `addr` containing an address.
+The second argument `data` is the data which is to be stored.
+The 3rd and 4th arguments can contain the following values.
+
+| **Memory Order**  | **Value** | **Summary**                                                                                        |
+| ------------------| --------- | -------------------------------------------------------------------------------------------------- |
+| Relaxed              | 0         | No constraints imposed on other reads or writes, only this operation's atomicity is guaranteed. |
```
WonderfulVoid: I don't know if the ACLE has some precedent here but otherwise, I would expect us to use GCC's `__ATOMIC_SEQ_CST`, `__ATOMIC_RELEASE` and `__ATOMIC_RELAXED` names and values. But perhaps the actual values are not stable and thus not suitable for ACLE?

main/acle.md
```diff
@@ -1826,6 +1827,13 @@ The `__ARM_FEATURE_SYSREG128` macro can only be implemented in the AArch64
 execution state. Intrinsics for the use of these instructions are specified in
 [Special register intrinsics](#special-register-intrinsics).
 
+### Producer-consumer data placement hints
+
+`__ARM_FEATURE_PCDPHINT` is defined to `1` if there is hardware
```
WonderfulVoid: Isn't `__ARM_FEATURE_PCDPHINT` defined (to 1) if there is **compiler** support? Hardware support is something else that can only be detected at runtime.

main/acle.md
```diff
@@ -4746,6 +4764,39 @@ stored to memory is modified by replacing the low 32 bits of
...
+These intrinsics provide an atomic store intrinsic, which will
+make use of the `STSHH` hint instruction immediately followed by the
+associated store instruction.
+This intrinsic is available when `__ARM_FEATURE_PCDPHINT` is defined.
+
+To access these intrinsics, `<arm_acle.h>` should be included.
+
+``` c
+  // Variants are also available for:
+  // _s8, _u16, _s16, _u32, _s32, _u64, _s64, _u128,
```
WonderfulVoid: We don't need actual prototypes for these other variants? Or that is handled in some other (magic) way?

main/acle.md
```diff
@@ -4746,6 +4764,39 @@ stored to memory is modified by replacing the low 32 bits of
...
+```
+
+The first argument in these intrinsics is a pointer `addr` containing an address.
+The second argument `data` is the data which is to be stored.
+The 3rd and 4th arguments can contain the following values.
+
+| **Memory Order**     | **Value** | **Summary**                                                                                     |
+| ---------------------| --------- | ----------------------------------------------------------------------------------------------- |
+| __ATOMIC_RELAXED     | 0         | No constraints imposed on other reads or writes, only this operation's atomicity is guaranteed. |
+| __ATOMIC_RELEASE     | 3         | No reads or writes in the current thread can be reordered after this store.                     |
```
WonderfulVoid: For completeness, I think we should also support __ATOMIC_SEQ_CST. Even if that likely will generate the same machine code (e.g. STLR).

main/acle.md
```diff
@@ -4746,6 +4764,39 @@ stored to memory is modified by replacing the low 32 bits of
...
+                      int ret); /* Retention Policy */
+```
+
+The first argument in these intrinsics is a pointer `addr` containing an address.
+The second argument `data` is the data which is to be stored.
+The 3rd and 4th arguments can contain the following values.
+
+| **Memory Order**     | **Value** | **Summary**                                                                                     |
+| ---------------------| --------- | ----------------------------------------------------------------------------------------------- |
+| __ATOMIC_RELAXED     | 0         | No constraints imposed on other reads or writes, only this operation's atomicity is guaranteed. |
```
WonderfulVoid: Do we need to specify the actual numeric values? Are we also specifying an ABI here?

main/acle.md
```diff
@@ -4746,6 +4764,39 @@ stored to memory is modified by replacing the low 32 bits of
...
+associated store instruction.
+This intrinsic is available when `__ARM_FEATURE_PCDPHINT` is defined.
+
+To access these intrinsics, `<arm_acle.h>` should be included.
+
+``` c
+  // Variants are also available for:
+  // _s8, _u16, _s16, _u32, _s32, _u64, _s64, _u128,
+  // _mf8, _bf16, _f16, _f32, _f64
+  void __arm_stshh_u8(uint8_t *addr,
```
Wilco1: Wouldn't it be better to define the store as type generic like __atomic_store_n?

Do we know the instruction sequences for the float and 128-bit variants? Do they rely on other extensions?

main/acle.md
```diff
@@ -1826,6 +1827,12 @@ The `__ARM_FEATURE_SYSREG128` macro can only be implemented in the AArch64
 execution state. Intrinsics for the use of these instructions are specified in
 [Special register intrinsics](#special-register-intrinsics).
 
+### Producer-consumer data placement hints
+
+`__ARM_FEATURE_PCDPHINT` is defined to `1` if the producer-consumer
+data placement hints (FEAT_PCDPHINT) instructions and their associated
+intrinsics are available on the target.
```
ktkachov, line 1845: IMO as these are hint instructions the developer may want to use them on older -march targets, so the __ARM_FEATURE_PCDPHINT should only signify whether the compiler version in question supports the intrinsics, but the intrinsics themselves should be always available (presumably the hint instructions execute as NOPs on older HW?)

ktkachov: Though reading a bit more, the user may not want to execute a PRFM instruction unless they know FEAT_PCDPHINT is present, so maybe it is prudent to gate these on hardware support. I don't insist on it.

main/acle.md
ricbal02: Upon discussion, I think an approach like __atomic_store_n would be a better implementation

main/acle.md
ricbal02: Quite possibly not, assuming these are well known values, we can maybe just omit that information

main/acle.md
ricbal02: I'm inclined to agree, the more I look into the implementation aspect of this, it seems that providing support for all memory ordering would be a negligible effort

main/acle.md
ricbal02: The ACLE seems to have precedence for just defining one prototype and allowing common sense to figure out how the other variants would be written. In any case, from Wilco's comment below, A generic approach like atomic_store_n seems favourable

main/acle.md
```diff
@@ -4746,6 +4764,32 @@ stored to memory is modified by replacing the low 32 bits of
...
+``` c
+  void __arm_atomic_store_with_stshh(type *ptr,
+                                     type data,
+                                     int mem, /* Memory order */
+                                     int ret); /* Retention Policy */
+```
+
+The first argument in these intrinsics is a pointer `ptr` which is the location to store to.
+The second argument `data` is the data which is to be stored.
+The third argument `mem` can be any of the 6 memory order variables from the C++11 standard.
```
Wilco1: __atomic_store only accepts __ATOMIC_RELAXED, __ATOMIC_SEQ_CST, and __ATOMIC_RELEASE. I don't see why allowing anything else would be useful.

ricbal02: Good catch, thanks

main/acle.md
```diff
@@ -4746,6 +4764,33 @@ stored to memory is modified by replacing the low 32 bits of
...
+  void __arm_atomic_store_with_stshh(type *ptr,
+                                     type data,
+                                     int mem, /* Memory order */
+                                     int ret); /* Retention Policy */
+```
+
+The first argument in these intrinsics is a pointer `ptr` which is the location to store to.
+The second argument `data` is the data which is to be stored.
+The third argument `mem` can be one of 3 memory ordering variables from the C++11 standard supported by atomic_store:
+__ATOMIC_RELAXED, __ATOMIC_SEQ_CST, and __ATOMIC_RELEASE.
```
tommat01, line 4854: I don't think these are in the C++11 standard with this spelling, these are [GCC builtins](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html). Also, it should probably reference C rather than C++.

main/acle.md
```diff
@@ -4746,6 +4764,33 @@ stored to memory is modified by replacing the low 32 bits of
...
+make use of the `STSHH` hint instruction immediately followed by the
+associated store instruction.
+This intrinsic is available when `__ARM_FEATURE_PCDPHINT` is defined.
+
+To access these intrinsics, `<arm_acle.h>` should be included.
+
+``` c
+  void __arm_atomic_store_with_stshh(type *ptr,
+                                     type data,
+                                     int mem, /* Memory order */
```
tommat01: Should the type of this be `memory_order` like in C11?

main/acle.md
```diff
@@ -1826,6 +1827,12 @@ The `__ARM_FEATURE_SYSREG128` macro can only be implemented in the AArch64
 execution state. Intrinsics for the use of these instructions are specified in
 [Special register intrinsics](#special-register-intrinsics).
 
+### Producer-consumer data placement hints
```
mgabka, line 1841: could you please specify at what quality level this proposal is being added?

main/acle.md
ricbal02: Sorry for the delay in response, You're correct about the hints operating as NOPs on older hardware so I would be inclined to agree on that point. I think that perhaps for the avoidance of doubt on the user's part having the PRFM gated would be beneficial

main/acle.md
ricbal02: Yes good catch, I noticed these same builtins in LLVM and assumed from there. I'll look into this

main/acle.md
ricbal02: Agreed

main/acle.md
```diff
@@ -4746,6 +4764,33 @@ stored to memory is modified by replacing the low 32 bits of
...
+
+These intrinsics provide an atomic store intrinsic, which will
+make use of the `STSHH` hint instruction immediately followed by the
+associated store instruction.
+This intrinsic is available when `__ARM_FEATURE_PCDPHINT` is defined.
+
+To access these intrinsics, `<arm_acle.h>` should be included.
+
+``` c
+  void __arm_atomic_store_with_stshh(type *ptr,
```
Wilco1, line 4845: There should be some explanation that the function is type generic and supports scalar integer types from 8-64 bits.

main/acle.md
ricbal02: LLVM seems to use the same wording for it's existing atomic_store, so I think this naming is the way to go, however I will change the wording of the explanation.

main/acle.md
ricbal02: Agreed

jthackray (approved):
LGTM

vhscampos (approved):
LGTM. Thank you

sallyarmneale (commented):
Looks good to me

PR #407, opened 2025-08-20
Add intrinsics for the new FP conversions introduced by the 2024 dpISA

amilendra:
FEAT_FPRCVT adds 4 new variants for each FCVTAS, FCVTAU, FCVTMS, FCVTMU, FCVTNS, FCVTNU, FCVTPS, FCVTPU, FCVTZS, and FCVTZU instruction. 1) Half Precision to 32-bit
2) Half Precision to 64-bit
3) Single Precision to 64-bit
4) Double Precision to 32-bit

For the Single Precision to 64-bit and Double Precision to 32-bit variants, this patch adds two new intrinsics, that reduce to
- Single Precision to 64-bit : <INST> Dd,Sn
- Double Precision to 32-bit : <INST> Sd,Dn

The intrinsics for conversions from Half Precision are already defined. However they are documented as reducing to the incorrect instruction format; <INST> Hd,Hn, so this patch fixes them to be
- Half Precision to 32-bit   : <INST> Sd,Hn
- Half Precision to 64-bit   : <INST> Dd,Hn

---
name: Pull request

---

If this PR is about a bugfix:

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.


main/acle.md
```diff
@@ -2207,6 +2210,13 @@ ACLE intrinsics are available. This implies that `__ARM_FEATURE_SM4` and
 floating-point absolute minimum and maximum instructions (FEAT_FAMINMAX)
 and if the associated ACLE intrinsics are available.
 
+### Floating-point to-from integer conversion intrinsics
+
+`__ARM_FEATURE_FPRCVT` is defined to `1` if there is hardware
+support for floating-point to integer and integer to floating-point
+conversion instructions. This includes the instructions FCVTAS, FCVTAU, FCVTMS,
```
Lukacma: I think this is bit misleading. we have floating point conversions even without this feature. This feature just enables us to do it within SIMD&FP reg file. I would also adjust the header to smth like:

SIMD&FP registers only floating-point conversions extension

or maybe just FPRCVT extension.

Listing all instructions here is also not necessary I think.

main/acle.md
```diff
@@ -2590,6 +2600,7 @@ be found in [[BA]](#BA).
 | [`__ARM_FEATURE_FP8DOT2`](#modal-8-bit-floating-point-extensions)                                                                                       | Modal 8-bit floating-point extensions                                                              | 1           |
 | [`__ARM_FEATURE_FP8DOT4`](#modal-8-bit-floating-point-extensions)                                                                                       | Modal 8-bit floating-point extensions                                                              | 1           |
 | [`__ARM_FEATURE_FP8FMA`](#modal-8-bit-floating-point-extensions)                                                                                        | Modal 8-bit floating-point extensions                                                              | 1           |
+| [`__ARM_FEATURE_FPRCVT`](#floating-point-to-from-integer-conversion-intrinsics)                                                                         | Floating-point to/from integer conversion intrinsics                                               | 1           |
```
Lukacma: Same as above. This description is bit misleading in my opinioon.

tools/intrinsic_db/advsimd.csv
```diff
@@ -4083,35 +4103,35 @@ float16_t vcvth_f16_u16(uint16_t a)	a -> Hn	UCVTF Hd,Hn	Hd -> result	A64
 float16_t vcvth_f16_u32(uint32_t a)	a -> Hn	UCVTF Hd,Hn	Hd -> result	A32/A64
 float16_t vcvth_f16_u64(uint64_t a)	a -> Hn	UCVTF Hd,Hn	Hd -> result	A64
 int16_t vcvth_s16_f16(float16_t a)	a -> Hn	FCVTZS Hd,Hn	Hd -> result	A64
-int32_t vcvth_s32_f16(float16_t a)	a -> Hn	FCVTZS Hd,Hn	Hd -> result	A32/A64
-int64_t vcvth_s64_f16(float16_t a)	a -> Hn	FCVTZS Hd,Hn	Hd -> result	A64
+int32_t vcvth_s32_f16(float16_t a)	a -> Hn	FCVTZS Sd,Hn	Sd -> result	A32/A64
```
Lukacma: I am not usre if A32 should stay here. I think these instructions are A64 only but I might be wrong

ktkachov:
As these instructions operate only on scalar values should they be defined in the arm_acle.h header instead of arm_neon.h?

main/acle.md
```diff
@@ -2207,6 +2208,13 @@ ACLE intrinsics are available. This implies that `__ARM_FEATURE_SM4` and
 floating-point absolute minimum and maximum instructions (FEAT_FAMINMAX)
 and if the associated ACLE intrinsics are available.
 
+### FPRCVT extension
+
+`__ARM_FEATURE_FPRCVT` is defined to `1` if there is hardware
+support for floating-point to/from integer convertion instructions
```
Lukacma:
```suggestion
support for floating-point to/from integer conversion instructions
```

main/acle.md
```diff
@@ -2207,6 +2208,13 @@ ACLE intrinsics are available. This implies that `__ARM_FEATURE_SM4` and
...
+with only scalar SIMD&FP register operands and results having
```
Lukacma:
```suggestion
with only scalar SIMD&FP register operands and results, and with
```

tools/intrinsic_db/advsimd.csv
```diff
@@ -4083,35 +4103,35 @@ float16_t vcvth_f16_u16(uint16_t a)	a -> Hn	UCVTF Hd,Hn	Hd -> result	A64
 float16_t vcvth_f16_u32(uint32_t a)	a -> Hn	UCVTF Hd,Hn	Hd -> result	A32/A64
 float16_t vcvth_f16_u64(uint64_t a)	a -> Hn	UCVTF Hd,Hn	Hd -> result	A64
 int16_t vcvth_s16_f16(float16_t a)	a -> Hn	FCVTZS Hd,Hn	Hd -> result	A64
-int32_t vcvth_s32_f16(float16_t a)	a -> Hn	FCVTZS Hd,Hn	Hd -> result	A32/A64
-int64_t vcvth_s64_f16(float16_t a)	a -> Hn	FCVTZS Hd,Hn	Hd -> result	A64
+int32_t vcvth_s32_f16(float16_t a)	a -> Hn	FCVTZS Sd,Hn	Sd -> result	A64
```
Lukacma: This is bit tricky situation here. All these intrinsics will lower to either new SIMD&FP conversion or the old GPR variants  depending on the context. So saying that they will always lower to SIMD&FP variant is wrong as can be seen here https://godbolt.org/z/nrae45Ea8. I am not sure how to express that here. Maybe  smth like this:


```suggestion
int32_t vcvth_s32_f16(float16_t a)	a -> Hn	FCVTZS Sd/Wd, Hn	Sd/Wd -> result	A64
```

tools/intrinsic_db/advsimd.csv
```diff
@@ -1168,15 +1168,25 @@ int32x4_t vcvtaq_s32_f32(float32x4_t a)	a -> Vn.4S	FCVTAS Vd.4S,Vn.4S	Vd.4S -> r
 uint32x2_t vcvta_u32_f32(float32x2_t a)	a -> Vn.2S	FCVTAU Vd.2S,Vn.2S	Vd.2S -> result	A32/A64
 uint32x4_t vcvtaq_u32_f32(float32x4_t a)	a -> Vn.4S	FCVTAU Vd.4S,Vn.4S	Vd.4S -> result	A32/A64
 int32_t vcvts_s32_f32(float32_t a)	a -> Sn	FCVTZS Sd,Sn	Sd -> result	A64
+int64_t vcvts_s64_f32(float32_t a)	a -> Sn	FCVTZS Dd,Sn	Dd -> result	A64
```
Lukacma: Is there a reason you didn't include GPR variants here ?

amilendra: My reasoning was that the fact that there were no existing conversion intrinsics that lowered to GPRs (https://developer.arm.com/documentation/ddi0602/2025-06/SIMD-FP-Instructions/FCVTZS--scalar--integer---Floating-point-convert-to-signed-integer--rounding-toward-zero--scalar--) we would only need to implement the intrinsics that lower to Neon registers. Thinking a bit more I guess it is better to extend the proposal to intrinsics that lower to both GP and Neon registers?

Lukacma (approved):
LGTM

tools/intrinsic_db/advsimd.csv
AlfieRichardsArm: I dont know why this was marked as resolved? We shouldnt remove the AArch32 variants right?

Lukacma: I don't know for sure, but from what I could find the only conversion instruction I could find in A32 is [this one](https://developer.arm.com/documentation/100076/0200/a32-t32-instruction-set-reference/advanced-simd-instructions--32-bit-/vcvt--from-floating-point-to-integer-with-directed-rounding-modes-?lang=en) which operates on 32 bit values only. So maybe only intrinsics for this variant should exist in A32 and rest should be A64 only ?

CarolineConcatto (approved):

tools/intrinsic_db/advsimd.csv
```diff
@@ -4083,35 +4103,35 @@ float16_t vcvth_f16_u16(uint16_t a)	a -> Hn	UCVTF Hd,Hn	Hd -> result	A64
 float16_t vcvth_f16_u32(uint32_t a)	a -> Hn	UCVTF Hd,Hn	Hd -> result	A32/A64
 float16_t vcvth_f16_u64(uint64_t a)	a -> Hn	UCVTF Hd,Hn	Hd -> result	A64
 int16_t vcvth_s16_f16(float16_t a)	a -> Hn	FCVTZS Hd,Hn	Hd -> result	A64
-int32_t vcvth_s32_f16(float16_t a)	a -> Hn	FCVTZS Hd,Hn	Hd -> result	A32/A64
-int64_t vcvth_s64_f16(float16_t a)	a -> Hn	FCVTZS Hd,Hn	Hd -> result	A64
+int32_t vcvth_s32_f16(float16_t a)	a -> Hn	FCVTZS Sd/Wd,Hn	Sd/Wd -> result	A64
+int64_t vcvth_s64_f16(float16_t a)	a -> Hn	FCVTZS Dd/Xd,Hn	Dd/Xd -> result	A64
 uint16_t vcvth_u16_f16(float16_t a)	a -> Hn	FCVTZU Hd,Hn	Hd -> result	A64
-uint32_t vcvth_u32_f16(float16_t a)	a -> Hn	FCVTZU Hd,Hn	Hd -> result	A32/A64
```
AlfieRichardsArm, line 4089: It seems here the A32 variant is removed? Why is this? I dont think we should do that.

AlfieRichardsArm (changes requested):

tools/intrinsic_db/advsimd.csv
Lukacma: > I don't know for sure, but from what I could find the only conversion instruction I could find in A32 is [this one](https://developer.arm.com/documentation/100076/0200/a32-t32-instruction-set-reference/advanced-simd-instructions--32-bit-/vcvt--from-floating-point-to-integer-with-directed-rounding-modes-?lang=en) which operates on 32 bit values only. So maybe only intrinsics for this variant should exist in A32 and rest should be A64 only ?

AlfieRichardsArm: There are more conversion instructions [here](https://developer.arm.com/documentation/100076/0200/a32-t32-instruction-set-reference/floating-point-instructions--32-bit-/vcvt--between-floating-point-and-integer-?lang=en).

But regardless, if we currently support these intrinsics for AAarch32, and have done for some time, removing them is a  breaking change that would need much wider discussion?

As I understand it, this extension adds support for existing conversions in AArch64 but allows the integer part to remain in the FP registers. It should have no effect on our AArch32 intrinsics so removing AArch32 intrinsics here seems wrong.

Lukacma: Okay. I don't think any compiler actually supports these intrinsics in aarch32 as their corresponding instructions don't exist, but I agree this change is unrelated to the core of the patch so I will remove it.

AlfieRichardsArm: Ah sorry I didn't realize these hadn't been implemented. Maybe we can get away with removing them then?

Lukacma: I am really fine going either way here.  @CarolineConcatto  what do you think ?

AlfieRichardsArm (approved):
Okay after the clarification with @Lukacma this is okay with me either way

vhscampos (approved):
LGTM. Merging

neon_intrinsics/advsimd.md
```diff
@@ -12,7 +12,7 @@ toc: true
 ---
 
 <!--
-SPDX-FileCopyrightText: Copyright 2014-2024 Arm Limited and/or its affiliates <open-source-office@arm.com>
+SPDX-FileCopyrightText: Copyright 2014-2025 Arm Limited and/or its affiliates <open-source-office@arm.com>
```
sallyarmneale, line 15: Should this be updated to 2026?

neon_intrinsics/advsimd.md
```diff
@@ -107,7 +107,7 @@ for more information about Arm’s trademarks.
 
 ## Copyright
 
-* Copyright 2014-2024 Arm Limited and/or its affiliates <open-source-office@arm.com>
+* Copyright 2014-2025 Arm Limited and/or its affiliates <open-source-office@arm.com>
```
sallyarmneale, line 110: Should this be updated to 2026?

neon_intrinsics/advsimd.template.md
```diff
@@ -12,7 +12,7 @@ toc: true
 ---
 
 <!--
-SPDX-FileCopyrightText: Copyright 2014-2024 Arm Limited and/or its affiliates <open-source-office@arm.com>
+SPDX-FileCopyrightText: Copyright 2014-2025 Arm Limited and/or its affiliates <open-source-office@arm.com>
```
sallyarmneale, line 15: Should this be updated to 2026?

neon_intrinsics/advsimd.template.md
```diff
@@ -107,7 +107,7 @@ for more information about Arm’s trademarks.
 
 ## Copyright
 
-* Copyright 2014-2024 Arm Limited and/or its affiliates <open-source-office@arm.com>
+* Copyright 2014-2025 Arm Limited and/or its affiliates <open-source-office@arm.com>
```
sallyarmneale, line 110: Should this be updated to 2026?

sallyarmneale (commented):
Few changes required for copyright years.

PR #413, opened 2025-09-30
Restrict FMV to not be combined with other mangling/cloning extensions

AlfieRichardsArm:
Function Multi Versioning should not be used alongside any C/C++ extensions
or attributes that affect function name mangling, or that clone functions.

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -465,6 +465,7 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 
 * Added feature test macro for FEAT_SSVE_FEXPA.
 * Added feature test macro for FEAT_CSSC.
+* Added restrictions of Function Multi Versioning's use with other extensions.
 
 ### References
 
@@ -2788,6 +2789,8 @@ The following attributes trigger the multi version code generation:
   * as a function annotated with `target_version("default")`,
   * or, as a function annotated with `target_clones(...)` where one
     of the versions is `default`.
+* Using any C/C++ extensions or attributes that affect function name mangling,
+  or that clone functions alongside Function Multi Versioning is not supported.
 
 The attribute `__attribute__((target_version("<target version string>")))` expresses the
 following:


main/acle.md
```diff
@@ -2788,6 +2789,8 @@ The following attributes trigger the multi version code generation:
   * as a function annotated with `target_version("default")`,
   * or, as a function annotated with `target_clones(...)` where one
     of the versions is `default`.
+* Function multi-versioning may not be used with any C/C++ extensions or
```
vhscampos: I suggest you use _must_ instead of _may_ to avoid ambiguity of intention. That is if I read correctly the meaning of the PR.

rockdreamer: I think that, if the intention is to warn users that the outcome of mixing is likely bad, both terms serve the purpose, "must" making it clearer for non-British speakers.
If the intention is to say that conforming toolchains can or should actively disallow such mixing, then I'd make that more explicit. LGTM :)

AlfieRichardsArm: Changed to "must", thanks!

main/acle.md
```diff
@@ -2788,6 +2789,8 @@ The following attributes trigger the multi version code generation:
   * as a function annotated with `target_version("default")`,
   * or, as a function annotated with `target_clones(...)` where one
     of the versions is `default`.
+* Function multi-versioning must not be used with any C/C++ extensions or
+  attributes which affect function name mangling, or that clone functions.
```
labrinea: This seems vague/generic. Is that intentional? Which specific extensions/attributes did you have in mind?

AlfieRichardsArm: In particular I was thinking of:
- GCC `asm` declarations
- GCC `rename` pragmas
- OpenMP `#pragma omp declare simd`

But I imagine there are more.

I am trying to be generic enough to avoid all extensions that effect mangling or clone functions, as I think they all would require a lot of consideration as to how they would work.

labrinea: From what I remember @jroelofs gave an example of `asm` being used to refer to a specific version here https://github.com/llvm/llvm-project/issues/84094 as a workaround for not supporting this functionality formally under a different syntax. The reasoning was to be able to fallback to lower priority versions, but I suppose the ability of the user to override priorities may aleviate this issue?

AlfieRichardsArm: Ah sorry, that was misleading

I meant asm used like:
```c
int fn (int) asm("different_func_name");

int fn [[gnu::target_version("sve")]] (int) { return 1; }
```

jroelofs: > The reasoning was to be able to fallback to lower priority versions, but I suppose the ability of the user to override priorities may aleviate this issue?

No, that doesn't cover the use-case I had in mind. Often the presence or absence of support for the hardware feature doesn't paint the full picture on whether or not we may want to select a given version. Take SME for example: for very small problem sizes, it sometimes makes sense to fall back on a NEON implementation, even with the SME ISA is available. For that specific case, it may make sense to call the NEON implementation _from_ the SME implementation, depending on the problem size, which the resolver cannot bake in at load time (nor lazily), since the problem size is often dynamic.

re: what Alfie is actually worried about: you may want to call out `__attribute__((overriddeable))` specifically, in addition to `asm()` and the others.

jroelofs: Why do we want to forbid this though? Might make more sense to define it as UB, or as implementation defined behavior, and then let vendors/platforms decide whether it makes sense to support it?

AlfieRichardsArm: I leant towards explicitly forbidding it as it leaves us the space to have specified behaviour in the future, or to change/introduce other FMV behaviour that might collide poorly with any implementations choice here.

AlfieRichardsArm: @jroelofs is overrideable a clang thing? I cant find any reference to it in the [docs](https://clang.llvm.org/docs/AttributeReference.html)?

labrinea: @smithp35 may have a suggestion about the exact wording. My thinking is how about we say something like "Using any C/C++ extensions or attributes which affect function name mangling, or that clone functions alongside FMV is not supported."?

AlfieRichardsArm: Ooh yeah I prefer that wording, will update to use that.

smithp35: Not sure I'm still needed. Reading the sentence is clear to me. I recommend using "that" rather than "which" as in American English "which" implies supplemental, non-essential information. In British English they are more interchangeable.

main/acle.md
```diff
@@ -2788,6 +2789,8 @@ The following attributes trigger the multi version code generation:
   * as a function annotated with `target_version("default")`,
   * or, as a function annotated with `target_clones(...)` where one
     of the versions is `default`.
+* Using any C/C++ extensions or attributes that affect function name mangling,
+  or that clone functions alongside FMV is not supported
```
labrinea: I don't think we state FMV as an acronym, so better use Function Multi Versioning instead.

labrinea: Also a fullstop is missing from the sentence.

AlfieRichardsArm: Fixed!

main/acle.md
AlfieRichardsArm: Fixed, thank you @smithp35

main/acle.md
labrinea: I am not seeing my points being addressed in the latest revision.

AlfieRichardsArm: Ah sorry, didn't `git add` 🤦. Updated now!

labrinea: I am seeing another mention of the acronym in a previous bullet point without actually defining it earlier. Perhaps we can add it in a parenthesis or remove the other mention of FMV.

AlfieRichardsArm: Yeah good point. I will address that in another PR just for the sake of keeping this clean.

AlfieRichardsArm: See: #417

labrinea: So far in the document it is being mentioned as Function Multi Versioning. The ABI document https://github.com/ARM-software/abi-aa/blob/main/sysvabi64/sysvabi64.rst#function-multi-versioning mentions it as `Function Multi-versioning (FMV)`. The learning path https://learn.arm.com/learning-paths/cross-platform/function-multiversioning/semantics `function multiversioning`. I am not sure what is more correct and whether it matters much. Perhaps it's best to keep the docs aligned, or at least all references within a doc the same. I am not opposed to introducing the FMV acronym in ACLE. I'd say use capital letters here (Function Multi Versioning) and we can amend later.

AlfieRichardsArm: Ah yes sorry about that, I agree consistency is most important. Changed to use `Function Multi Versioning`.

labrinea (approved):

main/acle.md
```diff
@@ -4,7 +4,7 @@ version: 2025Q3
 date-of-issue: 14 November 2025
 # LaTeX specific variables
 copyright-text: "Copyright: see section \\texorpdfstring{\\nameref{copyright}}{Copyright}."
```
sallyarmneale, line 6: Check that the copyright years variable is updated for 2026

sallyarmneale (commented):
copyright check required

PR #419, opened 2025-10-15
Improve documentation for VMLA/VMLS intrinsics for floats

amilendra:
The clang and GCC implementations for the VMLA/VMLS intrinsics for floats are the same except for the -ffp-contract=on case where GCC: uses VMUL+VADD/VSUM
Clang: uses FMLA/FMLS (Fused MLA/MLS)

Non-float VMLA/VMLS intrinsics reduce to MLA/MLS in both implementations.

Previous ACLE releases (e.g. 2.1) made it clear that the VMLA/VMLS intrinsics for floats are implementation defined, but this information seem to have got lost in later releases.

So redocument the implementation-defined behaviour.

---
name: Pull request

---

If this PR is about a bugfix:

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.


amilendra:
> > The current documentation say's nothing about contraction. It says "This intrinsic compiles to the following instructions: RESULT[I] = a[i] + (b[i] * c[i])"? When I build this with -Ofast I get contraction.
> 
> Yes, this was a bad specification, and it happened because when Armv8-a was added, the decision was made to switch all the "instructions" in the standard to use Arvm8-a instructions. This of course doesn't work when Armv8-a didn't have a matching instruction.
> 
> But note how the `vfma` one explicitly said it's a fused operation. I think they missed adding `unfused` to the `vmla` one because they didn't keep in mind `fp-contract` when the example was written.
> 
> This compatibility with Armv7 however is the reason why many duplicate intrinsics exist, e.g. that's why we have both `vdup_n_s8` and `vmov_n_s8`.
> 
> > The specification has been poorly worded, with a difference in implementation being the result. We can either capture this, ideally in a way that suggests ways to avoid such issues (i.e. don't use flags whose behaviour is not guaranteed to be consistent across architectures), or we can continue to leave readers to work it out for themselves.
> 
> I'm not saying that the documentation doesn't need updating. But I don't agree in updating it to be "implementation defined" because that was never the intention of it. But things got lost during the way.
> 
> Note that GCC had this fused behavior before GCC 12. We explicitly took the decision to fix it due to complaints of people porting code. https://godbolt.org/z/WcfzTdf3E so this changed.
> 
> We did so because of two clauses in the ACLE manual https://arm-software.github.io/acle/main/acle.html#undefined-behavior and https://arm-software.github.io/acle/main/acle.html#specification-of-advanced-simd-intrinsics where in this case the semantics comes from Armv7/Armv6 and prior.
> 
> The confusion comes from instructions that don't exist in Armv8 but intrinsic were still added to aid porting of older code.

@TamarChristinaArm 
Clearly this needs some more expertise than what I have and the clang implementation should be changed to match the gcc implementation which should be based on the spec changed. Would you mind taking this over and improving the specification?

paulwalker-arm:
There are no plans to change the clang implementation.  It is my understanding that if the `Final instruction sequence is implementation defined` phrasing is removed then the original objection is resolved?

amilendra:
> There are no plans to change the clang implementation. It is my understanding that if the `Final instruction sequence is implementation defined` phrasing is removed then the original objection is resolved?

I see. Thanks. I've removed that part.

neon_intrinsics/advsimd.md
```diff
@@ -393,40 +393,40 @@ The intrinsics in this section are guarded by the macro ``__ARM_NEON``.
...
+| <code>uint32x4_t <a href="https://developer.arm.com/architectures/instruction-sets/intrinsics/vmlaq_u32" target="_blank">vmlaq_u32</a>(<br>&nbsp;&nbsp;&nbsp;&nbsp; uint32x4_t a,<br>&nbsp;&nbsp;&nbsp;&nbsp; uint32x4_t b,<br>&nbsp;&nbsp;&nbsp;&nbsp; uint32x4_t c)</code>     | `a -> Vd.4S`<br>`b -> Vn.4S`<br>`c -> Vm.4S`    | `MLA Vd.4S,Vn.4S,Vm.4S`                                   | `Vd.4S -> result`  | `v7/A32/A64`              |
+| <code>float32x2_t <a href="https://developer.arm.com/architectures/instruction-sets/intrinsics/vmla_f32" target="_blank">vmla_f32</a>(<br>&nbsp;&nbsp;&nbsp;&nbsp; float32x2_t a,<br>&nbsp;&nbsp;&nbsp;&nbsp; float32x2_t b,<br>&nbsp;&nbsp;&nbsp;&nbsp; float32x2_t c)</code>   | `N/A`                                           | `RESULT[I] = vadd(a[i], vmul(b[i], c[i])) for i = 0 to 1` | `N/A`              | `v7/A32/A64`              |
```
paulwalker-arm: Given we're describing the operation using existing builtins, I suggest:
```
| `N/A` | result = vadd(a, vmul(b, c)) | `N/A` | `v7/A32/A64`
```

paulwalker-arm (approved):

tools/intrinsic_db/advsimd.csv
```diff
@@ -1572,8 +1572,8 @@ uint16x4_t vmls_n_u16(uint16x4_t a, uint16x4_t b, uint16_t c)	a -> Vd.4H;b -> Vn
 uint16x8_t vmlsq_n_u16(uint16x8_t a, uint16x8_t b, uint16_t c)	a -> Vd.8H;b -> Vn.8H;c -> Vm.H[0]	MLS Vd.8H,Vn.8H,Vm.H[0]	Vd.8H -> result	v7/A32/A64
 uint32x2_t vmls_n_u32(uint32x2_t a, uint32x2_t b, uint32_t c)	a -> Vd.2S;b -> Vn.2S;c -> Vm.S[0]	MLS Vd.2S,Vn.2S,Vm.S[0]	Vd.2S -> result	v7/A32/A64
 uint32x4_t vmlsq_n_u32(uint32x4_t a, uint32x4_t b, uint32_t c)	a -> Vd.4S;b -> Vn.4S;c -> Vm.S[0]	MLS Vd.4S,Vn.4S,Vm.S[0]	Vd.4S -> result	v7/A32/A64
-float32x2_t vmls_n_f32(float32x2_t a, float32x2_t b, float32_t c)	N/A	RESULT[I] = a[i] - (b[i] * c) for i = 0 to 1	N/A	v7/A32/A64
-float32x4_t vmlsq_n_f32(float32x4_t a, float32x4_t b, float32_t c)	N/A	RESULT[I] = a[i] - (b[i] * c) for i = 0 to 3	N/A	v7/A32/A64
+float32x2_t vmls_n_f32(float32x2_t a, float32x2_t b, float32_t c)	N/A	result = vsub(a, vmul(b, c))	N/A	v7/A32/A64
```
AlfGalf, line 1575: Minor suggestion (not a blocker) is this may make more sense to use the full builtin names? eg. `vsub_f32` here instead of `vsub` as that isn't the name of any intrinsic? (and similar for the other cases)

AlfGalf (approved):
LGTM other than a minor suggestion (not a blocker)

tools/intrinsic_db/advsimd.csv
amilendra: @paulwalker-arm do you think adding using  `vadd_f32`, `vsub_f32`, and `vmul_f32` instead of `vadd`, `vsub`, and `vmul` respectively would improve the documentation?

paulwalker-arm: It is an improvement to the clarity. However, I figure the shorthand is easily interpretable and thus the cost of specialising all the instances is not worth the cost. So I'm happy either way, but personally I wouldn't bother.

amilendra: Thanks both. Because everyone agrees with the current version, and the suggestion is a minor improvement, I'll go with the current version.

main/acle.md
```diff
@@ -474,6 +474,7 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * Added restrictions of Function Multi Versioning's use with other extensions.
 * Upgrade Function Multi Versioning to Release support level.
 * Removed _single from svmla_za16[_mf8]_vg2x1_fpm and svmla_za32[_mf8]_vg4x1_fpm.
+* Improve documentation for VMLA/VMLS intrinsics for floats.
```
sallyarmneale, line 477: Improved?

main/acle.md
```diff
@@ -474,6 +474,7 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * Added restrictions of Function Multi Versioning's use with other extensions.
 * Upgrade Function Multi Versioning to Release support level.
```
sallyarmneale, line 475: Upgraded?

sallyarmneale (commented):
minor changes

PR #423, opened 2025-12-01
Add intrinsic support for the range prefetch (RPRFM) instruction

kmclaughlin-arm:
---
name: Pull request

---

If this PR is about a bugfix:

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.


main/acle.md
```diff
@@ -3613,6 +3614,51 @@ values.
...
+| **Metadata**   | **Range**         | **Summary**                                                          |
+| -------------- | ----------------- | -------------------------------------------------------------------- |
+| Reuse Distance | 0 to 15           | Maximum number of bytes to be accessed before executing the          |
+|                |                   | next RPRFM instruction that specifies the same range. Values         |
+|                |                   | from 1 to 15 represent decreasing powers of two in the range         |
+|                |                   | 512MiB to 32KiB. A value of 0 indicates distance not known.          |
+|                |                   | Note: This value is ignored if a streaming prefetch is specified.    |
+| Stride         | -2MiB to +2MiB-1B | Number of bytes to advance the block address by after `Length`       |
+|                |                   | bytes have been accessed. Note: This value is ignored if Count is 0. |
+| Count          | 0 to 65535        | Number of blocks to be accessed, minus 1.                            |
```
mgabka: Perhaps it would be easier for the user if Count was defined as Number of data blocks to be accessed, that would be more natural as we always access at least 1 data block.

main/acle.md
```diff
@@ -3613,6 +3614,51 @@ values.
 | KEEP                 | 0         | Temporal fetch of the addressed location (that is, allocate in cache normally) |
 | STRM                 | 1         | Streaming fetch of the addressed location (that is, memory used only once)     |
 
+The following intrinsic is also available when `__ARM_FEATURE_RPRFM` is defined:
+
+``` c
+  void __rpld(/*constant*/ unsigned int /*access_kind*/,
```
paulwalker-arm: perhaps `__pld_range`?

main/acle.md
```diff
@@ -3613,6 +3614,51 @@ values.
...
+              /*constant*/ unsigned int /*retention_policy*/,
+              /*constant*/ unsigned int /*reuse distance*/,
+              /*constant*/ signed int   /*stride*/,
+              /*constant*/ unsigned int /*count*/,
+              /*constant*/ signed int   /*length*/,
```
paulwalker-arm: These parameters are not linked to the opcode and thus don't need to be constant?

I don't know the expected usage so perhaps making these variable is asking for trouble? In which case I think it's worth adding a variant where the metadata is an opaque value so at least there's a way to generate the prefetch for the non-constant use case without needing to use inline asm.

main/acle.md
```diff
@@ -3613,6 +3614,51 @@ values.
...
+
+The table below describes the ranges of the reuse distance, stride, count and length arguments.
+
+| **Metadata**   | **Range**         | **Summary**                                                          |
+| -------------- | ----------------- | -------------------------------------------------------------------- |
+| Reuse Distance | 0 to 15           | Maximum number of bytes to be accessed before executing the          |
+|                |                   | next RPRFM instruction that specifies the same range. Values         |
+|                |                   | from 1 to 15 represent decreasing powers of two in the range         |
+|                |                   | 512MiB to 32KiB. A value of 0 indicates distance not known.          |
+|                |                   | Note: This value is ignored if a streaming prefetch is specified.    |
```
paulwalker-arm: Similar to the change to `Count` it will be more user friendly to accept the real world value and let the compiler encode it.

main/acle.md
```diff
@@ -3613,6 +3614,51 @@ values.
...
+| Stride         | -2MiB to +2MiB-1B | Number of bytes to advance the block address by after `Length`       |
```
paulwalker-arm: `[-2MiB, +2MiB)`?

main/acle.md
```diff
@@ -3613,6 +3614,51 @@ values.
 | KEEP                 | 0         | Temporal fetch of the addressed location (that is, allocate in cache normally) |
 | STRM                 | 1         | Streaming fetch of the addressed location (that is, memory used only once)     |
 
+The following intrinsic is also available when `__ARM_FEATURE_RPRFM` is defined:
```
paulwalker-arm: Is the expectation for the builtin to fail compilation when this feature is not present? or be treated as a NOP?  The existing `_pld` documentation suggests the latter.

I guess it comes down to whether the builtin provides useful information that can be used even if RPRFM is not available (e.g. by emitting other prefetch instructions, or non-temporal accesses).

The reason I mention this is because if there's only a weak link to the underlying instruction then we'll not want to use a feature macro, but instead something that just says the builtin is available, like __ARM_PREFIX_RANGE.

main/acle.md
```diff
@@ -3613,6 +3614,51 @@ values.
...
+              /*constant*/ unsigned int /*retention_policy*/,
+              /*constant*/ unsigned int /*reuse distance*/,
+              /*constant*/ signed int   /*stride*/,
+              /*constant*/ unsigned int /*count*/,
+              /*constant*/ signed int   /*length*/,
+              void const volatile *addr);
+```
+
+Generates a data prefetch instruction from a range of addresses starting from a
+given base address. Locations within the specified address ranges are prefetched
```
paulwalker-arm:
```suggestion
Generates a data prefetch instruction for a range of addresses starting from a
given base address. Locations within the specified address ranges are prefetched
```

main/acle.md
```diff
@@ -3613,6 +3614,51 @@ values.
...
+| -------------- | ----------------- | -------------------------------------------------------------------- |
+| Reuse Distance | 0 to 15           | Maximum number of bytes to be accessed before executing the          |
+|                |                   | next RPRFM instruction that specifies the same range. Values         |
+|                |                   | from 1 to 15 represent decreasing powers of two in the range         |
+|                |                   | 512MiB to 32KiB. A value of 0 indicates distance not known.          |
+|                |                   | Note: This value is ignored if a streaming prefetch is specified.    |
+| Stride         | -2MiB to +2MiB-1B | Number of bytes to advance the block address by after `Length`       |
+|                |                   | bytes have been accessed. Note: This value is ignored if Count is 1. |
+| Count          | 1 to 65536        | Number of blocks to be accessed.                                     |
+| Length         | -2MiB to +2MiB-1B | Number of contiguous bytes to be accessed.                           |
```
paulwalker-arm: [-2MiB, +2MiB)?

main/acle.md
kmclaughlin-arm: I think the existing documentation above the `_pld` builtin applies here too, in that the range builtin should not fail when the feature is not present and is treated as a nop.

main/acle.md
kmclaughlin-arm: Removed `/*constant*/` from the metadata arguments, as there was no specific reason requiring them to be constant.

main/acle.md
kmclaughlin-arm: I've changed this to accept either 0 (for distance not known) or a value representing the number of bytes instead.

main/acle.md
paulwalker-arm: I doubt it'll be this easy.  By allowing variable arguments the compiler will then need to sanitise their values, which I think will be non-trivial and potentially result in adding a prefetch causing performance regressions due to the unexpected overhead.  This is why I suggested a separate builtin where the metadata is a single opaque value, because then the user can see the overhead, because they've had to compute it manually.

main/acle.md
```diff
@@ -3613,6 +3614,51 @@ values.
...
+
+The table below describes the ranges of the reuse distance, stride, count and length arguments.
+
+| **Metadata**   | **Range**           | **Summary**                                                          |
+| -------------- | ------------------- | -------------------------------------------------------------------- |
+| Reuse Distance | 0 or [2**15, 2**29] | Maximum number of bytes to be accessed before executing the          |
+|                |                     | next RPRFM instruction that specifies the same range. Values         |
+|                |                     | are powers of two representing the number of bytes in the range      |
+|                |                     | 32KiB to 512MiB. A value of 0 indicates distance not known.          |
+|                |                     | Note: This value is ignored if a streaming prefetch is specified.    |
```
paulwalker-arm: Do you know how accurate `Reuse Distance` needs to be? It says  "Maximum number of bytes" so I take that to mean you can access less.  I ask because perhaps this can be freeform, with the compiler selecting the lowest compatible value, or 0 for values above the maximum allowed.

kmclaughlin-arm: I haven't found anything to say that `ReuseDistance` needs to be accurate, so my understanding is also that it could access less than the number of bytes requested.

main/acle.md
mgabka: are you worried about extra spills which may cause changes to the memory and the provided values won't match the memory pattern the prefetched instruction is using?

paulwalker-arm: Not specifically, just any extra work required to construct the metadata that would be invisible to the user and thus might surprise them.

main/acle.md
paulwalker-arm: Does that mean the range information can be removed because the compiler can do the necessary rounding for free. Then just call out 0 and values over 512MiB are treated as unknown?

kmclaughlin-arm: I think the range information is still accurate though, as long as it's in the description that any value outside the range will use 0 for unknown distance?

paulwalker-arm: To me the range information is specifying the builtin's acceptable values for that parameter, with everything else resulting in a diagnostic.  With the compiler doing implicit rounding, all values are acceptable and so there is no range.  The summary can then document how values will be interpreted (e.g.  All values are rounded up to the nearest power-of-two in the range low - high, with the exception of ....).

kmclaughlin-arm: Ok, I think it makes sense to remove the range given that the rounding means there won't be any diagnostics shown for invalid values.

main/acle.md
AlfGalf: Yeah I think I agree with Paul, I think we should have a __ARM_PREFETCH_RANGE macro to tell the user if this part of the ACLE is currently implemented for the compiler and is available. Then it can be used for conditional compilation?

I think we could also have a __ARM_FEATURE_RPRFM macro, to indicate if the actual feat is enabled though?

AlfGalf (commented):
The intrinsic itself looks good to me, slight question about the macro though

main/acle.md
```diff
@@ -3613,6 +3614,79 @@ values.
...
+| **Retention Policy** | **Value** | **Summary**                                                                |
+| -------------------- | --------- | -------------------------------------------------------------------------- |
+| KEEP                 | 0         | Temporal fetch of the addressed location (that is, allocate in cache normally) |
+| STRM                 | 1         | Streaming fetch of the addressed location (that is, memory used only once)     |
+
+The table below describes the ranges of the reuse distance, stride, count and length arguments.
+
+| **Metadata**   | **Range**           | **Summary**                                                          |
+| -------------- | ------------------- | -------------------------------------------------------------------- |
+| Reuse Distance |                     | Maximum number of bytes to be accessed before executing the next     |
```
CarolineConcatto, line 3667: Should we also add the Range here for Reuse Distance?

kmclaughlin-arm: Hi @CarolineConcatto, the reason for removing the Reuse Distance range was discussed here: https://github.com/ARM-software/acle/pull/423#discussion_r2607459257

main/acle.md
kmclaughlin-arm: I've updated this to describe the `__ARM_PREFETCH_RANGE` macro.

@AlfGalf  Given that the `__ARM_FEATURE_RPRFM` macro should be used to query whether this is implemented, I'm not sure why we would also want to add `__ARM_FEATURE_RPRFM` if the instruction does not require something like `+rprfm`?

mgabka: I support @kmclaughlin-arm , the FEAT_RPRFM is OPTIONAL from Armv8.0, and not guarded by any dedicated feature flag,  moreover it looks like RPRFM is a preferred disassembly for PRFM with UXTW, as guarded by this test https://github.com/llvm/llvm-project/blob/f9a3076180bfa1ebffd518971eefd30939a5ced6/llvm/test/MC/AArch64/rprfm.s#L273 , so not sure what we would achieve with having __ARM_FEATURE_RPRFM.

main/acle.md
```diff
@@ -3613,6 +3614,80 @@ values.
...
+The `__ARM_PREFETCH_RANGE` macro can be used to test for the presence of the
+following range prefetch intrinsics:
+
+``` c
+  void __pldx_range(/*constant*/ unsigned int /*access_kind*/,
+                    /*constant*/ unsigned int /*retention_policy*/,
+                    /*constant*/ size_t       /*reuse distance*/,
+                    /*constant*/ signed int   /*stride*/,
+                    /*constant*/ unsigned int /*count*/,
+                    /*constant*/ signed int   /*length*/,
```
paulwalker-arm: Not a strongly held view, but after reading the description of `__pld_range` I do feel its "parameters" have a more natural order (i.e. length, count, stride, reuse distance). What do you think to using the same order for `__pldx_range`?

paulwalker-arm (commented):
I'm mostly happy, with one last query.

main/acle.md
kmclaughlin-arm: I think it would be better if the order is consistent for both of the intrinsics, and it makes sense to keep the ordering from bits 0-63 used by `__pld_range`.

main/acle.md
AlfieRichardsArm: Ah yeah, okay, agreed best to not have __ARM_FEATURE_RPRFM

AlfieRichardsArm (approved):
LGTM

paulwalker-arm (approved):

main/acle.md
```diff
@@ -3613,6 +3614,80 @@ values.
...
+|                |                     | RPRFM instruction that specifies the same range. All values are      |
+|                |                     | rounded up to the nearest power of 2 in the range 32KiB to 512MiB.   |
+|                |                     | Values exceeding the maximum of 512MiB will be represented by 0,     |
+|                |                     | indicating distance not known.                                       |
+|                |                     | Note: This value is ignored if a streaming prefetch is specified.    |
+
+``` c
+  void __pld_range(/*constant*/ unsigned int /*access_kind*/,
+                   /*constant*/ unsigned int /*retention_policy*/,
+                   unsigned long /*metadata*/,
```
AlfieRichardsArm: I realize this is quite late to request a change, sorry about that, but is it worth using uint64_t for metadata instead of unsigned long to avoid any issues with targets where long != 64 bits?

AlfieRichardsArm: Or long long, happy with either :)

kmclaughlin-arm: No problem, I think it makes sense to change it. Metadata is now using `uint64_t`.

AlfieRichardsArm (approved):
LGTM :)

main/acle.md
AlfieRichardsArm: Thanks!

vhscampos (approved):
LGTM too. Thank you

main/acle.md
```diff
@@ -476,6 +476,7 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * Removed _single from svmla_za16[_mf8]_vg2x1_fpm and svmla_za32[_mf8]_vg4x1_fpm.
 * Improve documentation for VMLA/VMLS intrinsics for floats.
```
sallyarmneale, line 477: Improved?

sallyarmneale (commented):
Minor query

PR #428, opened 2026-01-30
Add alpha support for 9.7 data processing intrinsics

rockdreamer:
---
name: Add alpha support for 9.7 data processing intrinsics
about: feature proposal.

---

This change adds intrinsics for the following architectural features:
- FEAT_F16F32DOT
- FEAT_F16F32MM
- FEAT_F16MM
- FEAT_SVE_B16MM
- FEAT_SVE2p3
- FEAT_SME2p3

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.


main/acle.md
```diff
@@ -13857,6 +13939,128 @@ Scalar index of first/last true predicate element (predicated).
...
+The functions in this section are defined by either the header file
+ [`<arm_sve.h>`](#arm_sve.h) or [`<arm_sme.h>`](#arm_sme.h)
+when `__ARM_FEATURE_SVE2p3` or `__ARM_FEATURE_SME2p3` is defined, respectively.
+
+#### ADDQP
+
+Add pairwise within quadword vector segments.
+
+``` c
+  // Variants are also available for _s16, _s32 and _s64
```
georges-arm: We should also have unsigned variants here.

main/acle.md
```diff
@@ -13857,6 +13939,128 @@ Scalar index of first/last true predicate element (predicated).
...
+  svint8_t svaddqp[_s8](svint8_t zn, svint8_t zm);
+  ```
+
+#### ADDSUBP
+
+Add subtract pairwise.
+
+``` c
+  // Variants are also available for _s16, _s32 and _s64
```
georges-arm: We should also have unsigned variants here?

main/acle.md
```diff
@@ -13857,6 +13939,128 @@ Scalar index of first/last true predicate element (predicated).
...
+  //           f32[_s16], f64[_s32],
+  // f16[_u8], f32[_u16], f64[_u32]
+  
+  svfloat16_t svcvtt_f16[_s8](svint8_t zn);
+  svfloat16_t svcvtb_f16[_s8](svint8_t zn);
+```
+
+#### SDOT, UDOT (vectors)
+
+Integer dot-product (2-way, vectors).
```
georges-arm: "2-way" vs "two-way" are used inconsistently, pick one? Personal preference for the latter, but fine with either.

main/acle.md
```diff
@@ -13857,6 +13939,128 @@ Scalar index of first/last true predicate element (predicated).
...
+``` c
+  // Variants are also available for
+  //           f32[_s16], f64[_s32],
+  // f16[_u8], f32[_u16], f64[_u32]
+  
+  svfloat16_t svcvtt_f16[_s8](svint8_t zn);
+  svfloat16_t svcvtb_f16[_s8](svint8_t zn);
+```
+
+#### SDOT, UDOT (vectors)
```
georges-arm: There should already be a section for the _s32_s16 etc variant of this added with `FEAT_SVE2p1` somewhere else in this doc, should this be merged into that?

main/acle.md
```diff
@@ -13857,6 +13939,128 @@ Scalar index of first/last true predicate element (predicated).
...
+                                uint64_t imm_idx);
+  ```
+
+#### SQSHRN, UQSHRN
+
+Multi-vector saturating shift right narrow and interleave.
+
+``` c
+  // Variants are also available for 
+  //              _u16[_u32_x2]
```
georges-arm: Rework this so that signed/unsigned are on their own lines instead of grouping by 8/16, then we can fix the alignment here?

main/acle.md
```diff
@@ -13857,6 +13939,128 @@ Scalar index of first/last true predicate element (predicated).
...
+  // Variant for _u8[_s16_x2] is also available.
+  svuint16_t svqshrun[_n]_u16[_s32_x2](svint32x2_t zn, uint64_t imm);
+  ```
+
+#### SUBP
+
+Subtract pairwise.
+
+``` c
+  // Variants are also available for _s16, _s32 and _s64
```
georges-arm: This should have unsigned variants?

main/acle.md
rockdreamer: Thanks for reviewing, will fix

main/acle.md
rockdreamer: Will do

main/acle.md
rockdreamer: Agree that inconsistency is not nice but I'm following the wording in the Arm ARM where [SABAL](https://developer.arm.com/documentation/ddi0602/2025-12/SVE-Instructions/SABAL--Two-way-signed-absolute-difference-sum-and-accumulate-long-?lang=en) uses "Two-way" and SDOT "2-way".

main/acle.md
rockdreamer: Fixed, thanks for reviewing! :)

main/acle.md
rockdreamer: done

main/acle.md
rockdreamer: Done for both the indexed and vector SDOT/UDOT cases

neon_intrinsics/advsimd.md
```diff
@@ -6213,3 +6217,16 @@ The intrinsics in this section are guarded by the macro ``__ARM_NEON``.
...
+| <code>float32x4_t <a href="https://developer.arm.com/architectures/instruction-sets/intrinsics/vfdotq_f32_f16" target="_blank">vfdotq_f32_f16</a>(<br>&nbsp;&nbsp;&nbsp;&nbsp; float32x4_t r,<br>&nbsp;&nbsp;&nbsp;&nbsp; float16x8_t a,<br>&nbsp;&nbsp;&nbsp;&nbsp; float16x8_t b)</code>                                                       | `r -> Vd.4S`<br>`a -> Vn.8H`<br>`b -> Vm.8H`                     | `FDOT Vd.4S,Vn.8H,Vm.8H`       | `Vd.4S -> result` | `A64`                     |
+| <code>float32x2_t <a href="https://developer.arm.com/architectures/instruction-sets/intrinsics/vfdot_lane_f32_f16" target="_blank">vfdot_lane_f32_f16</a>(<br>&nbsp;&nbsp;&nbsp;&nbsp; float32x2_t r,<br>&nbsp;&nbsp;&nbsp;&nbsp; float16x4_t a,<br>&nbsp;&nbsp;&nbsp;&nbsp; float16x2_t b,<br>&nbsp;&nbsp;&nbsp;&nbsp; const int lane)</code>   | `r -> Vd.2S`<br>`a -> Vn.4H`<br>`b -> Vm.2H`<br>`0 <= lane <= 3` | `FDOT Vd.2S,Vn.4H,Vm.2H[lane]` | `Vd.2S -> result` | `A64`                     |
```
ricbal02: float16x4_t b for vfdot_lane and 
float16x8_t b for vfdot_laneq

main/acle.md
```diff
@@ -13281,7 +13373,8 @@ Set scalar to count from predicate-as-counter. ``vl`` is expected to be 2 or 4.
 Multi-vector dot-product (2-way)
 
 ``` c
-  // Variants are also available for _s32_s16 and _u32_u16
+  // Variants are also available for _s32_s16, _u32_u16
+  // and also for _s16_s8 and _u16_u8 if (__ARM_FEATURE_SVE2p3 __ARM || _FEATURE_SME2p3).
```
georges-arm: `__ARM` seems misplaced here?

main/acle.md
```diff
@@ -13293,7 +13386,8 @@ Multi-vector dot-product (2-way)
 Multi-vector dot-product (2-way)
 
 ``` c
-  // Variants are also available for _s32_s16 and _u32_u16
+  // Variants are also available for _s32_s16, _u32_u16
+  // and also for _s16_s8 and _u16_u8  if (__ARM_FEATURE_SVE2p3 __ARM || _FEATURE_SME2p3).
```
georges-arm: `__ARM` seems misplaced here?

main/acle.md
```diff
@@ -2425,6 +2472,21 @@ instructions and if the associated ACLE intrinsics are available.
 for the SVE 16-bit floating-point to 32-bit floating-point matrix multiply and add
 (FEAT_SVE_F16F32MM) instructions and if the associated ACLE intrinsics are available.
 
+##### Multiplication of 16-bit floating-point matrices (AdvSIMD)
+
+This section is in
+[**Alpha** state](#current-status-and-anticipated-changes) and might change or be
+extended in the future.
+
+`__ARM_FEATURE_F16F32MM` is defined if the NEON half-precision matrix multiply
```
georges-arm: `NEON` should be `Neon`. Same everywhere else.

main/acle.md
```diff
@@ -9594,6 +9661,15 @@ BFloat16 floating-point adjust exponent vectors.
 
 ### SVE2 floating-point matrix multiply-accumulate instructions.
 
+#### BFMMLA, FMMLA(non-widening)
```
georges-arm: Space before `(` ?

main/acle.md
```diff
@@ -13857,6 +13953,106 @@ Scalar index of first/last true predicate element (predicated).
...
+Use of this intrinsic if `svcntb() * 8 < 512` results in undefined behaviour.
+
+``` c
+  // Variants are also available for _u16_x2 and _f16_x2.
+  svint16_t svluti6_lane[_s16_x2](svint16x2_t table, svuint8_t indices, uint64_t imm_idx);
+  ```
+
+#### FCVTZSN, FCVTZUN
+
+Floating-point narrowing convert to interleaved integer, rounding toward zero
```
georges-arm: Missing full-stop (`.`) at end of this line, possibly others?

main/acle.md
```diff
@@ -13857,6 +13953,106 @@ Scalar index of first/last true predicate element (predicated).
...
+
+#### SABAL, UABAL
+
+Two-way absolute difference sum and accumulate long.
+
+``` c
+  // Variants are also available for
+  //            _s32[_s16], _s64[_s32],
+  // _u16[_u8], _u32[_u16], _u64[_u32].
+  svint16_t svaba_s16[_s8](svint16_t zda, svint8_t zn, svint8_t zm);
```
georges-arm: In other widening intrinsics we include `l` in the intrinsic name, so this should be `svabal` ?

main/acle.md
```diff
@@ -13857,6 +13953,106 @@ Scalar index of first/last true predicate element (predicated).
...
+
+#### FCVTZSN, FCVTZUN
+
+Floating-point narrowing convert to interleaved integer, rounding toward zero
+
+``` c
+  // Variants are also available for
+  //               _s16[_f32_x2], _s32[_f64_x2],
+  // _u8[_f16_x2], _u16[_f32_x2], _u32[_f64_x2].
+  svint8_t svcvt_s8[_f16_x2](svfloat16x2_t zn);
```
georges-arm: Other narrowing conversions have different naming - they include the suffix, e.g. `svcvtxnt`, so should this be `svsvcntzn` or similar?

main/acle.md
```diff
@@ -13996,6 +13999,17 @@ Floating-point narrowing convert to interleaved integer, rounding toward zero
...
 
+#### SABAL, UABAL
+
+Two-way absolute difference sum and accumulate long.
+
+``` c
+  // Variants are also available for
+  //            _s32[_s16], _s64[_s32],
+  // _u16[_u8], _u32[_u16], _u64[_u32].
+  svint16_t svaba_s16[_s8](svint16_t zda, svint8_t zn, svint8_t zm);
```
amilendra: Can't we do `svint16_t svaba[_s16_s8](svint16_t zda, svint8_t zn, svint8_t zm);`?
Or is `svint16_t svaba(svint16_t zda, svint8_t zn, svint8_t zm);` already defined elsewhere?

main/acle.md
abhishek-kaushik22: Adding to the previous comment, the current `svaba*` intrinsics have the naming scheme like `[_s16]` or `[_n_s16]` and don't include the second type (same in case of `svaba` but for widening intrinsics the half type).
Should this follow that as well?

neon_intrinsics/advsimd.md
rockdreamer: After some discussion, I'll use a similar shape to bfdot.

main/acle.md
rockdreamer: Sorted

main/acle.md
rockdreamer: I'll add a patch to fix all instances.

main/acle.md
rockdreamer: Had a look at all patches, caught another instance, thanks.

main/acle.md
rockdreamer: @georges-arm indeed, a look at [other svaba intrinsics](https://developer.arm.com/architectures/instruction-sets/intrinsics/#f:@navigationhierarchiesreturnbasetype=[%5Buint%5D]&q=svaba) gives me svabablb and svabalt.
There's also svcvtl from recent sme. I think the reason I used _s16[_s8] is that svcvtl is a unary conversion which does not alter the first argument, so it needs a disambiguator for the return type. On more careful thought, this is not necessary as we do pass zda.

I guess the reason I didn't go for a version with a final l is that I've been asking myself: what if we get an _s32_s8 version in the future? 
We could do `svint16_t svabal[_s16_s8](svint16_t zda, svint8_t zn, svint8_t zm);` and that would allow us to make more combinations.

Or maybe I'm overthinking is and `svint16_t svabal[_s16](svint16_t zda, svint8_t zn, svint8_t zm);` is good enough and if we need more we can always make a svaball.

I'm happy either way, do you have a preference? :)

The question I might add is: do we want to add a scalar version? For example `svint16_t svabal[_n_s16](svint16_t zda, svint8_t zn, int8_t zm);`. The reason I can see is that it would make these intrinsics have the same "shape" as  svabalb, svabalt, svmlalb etc. It would have to be synthesized, and we do this for svmlalb for example. I'm leaning towards this.

@abhishek-kaushik22 this text is lost in the vastness of acle.md "In a vector operation, the disambiguator `_n` indicates that the final operand is a scalar rather than a vector.".
So if we add the scalar version, you're looking at a ternary_long_opt_n in gcc shape parlance, otherwise I think you'll have to create a new shape.

main/acle.md
rockdreamer: I think the reason I used _s16[_s8] is that I took the hint from svcvtl, a unary conversion which does not alter the first argument. I've also commented above in George's review, let's concentrate the discussion there :).

main/acle.md
rockdreamer: I had a look at [other svcvt intrinsics](https://developer.arm.com/architectures/instruction-sets/intrinsics/#f:@navigationhierarchiesreturnbasetype=[%5Buint%5D,uint]&f:@navigationhierarchiessimdisa=[SVE2,SVE,SME%20and%20SME2]&q=svcvt&first=120) and could not find a conflict. So the simpler version made sense to me.
That being said, svcvtzn also makes sense. I'll switch to that.

main/acle.md
```diff
@@ -14648,6 +14687,36 @@ non-overloaded names to indicate which vector argument is a vector register pair
...
+[`<arm_sme.h>`](#arm_sme.h) when `__ARM_FEATURE_SME2p3` is defined to 1.
+
+#### LUTI6
+
+Lookup table read with 6-bit indices (16-bit)
+
+Use of this intrinsic if `svcntb() * 8 < 512` results in undefined behaviour.
+
+```c
+  // Variant are also available for: _u16, _f16 and _bf16.
```
jthackray: s/Variant/Variants/

main/acle.md
MartinWehking: Hi @rockdreamer `svcvtzn` sounds like a sensible naming choice.
I'm just wonderng if you're planning to update this in your patch, since I have put up a patch for LLVM already with the old naming choice:
https://github.com/llvm/llvm-project/pull/186807

Once this has been updated here, I would also rename it in my patch.

main/acle.md
rockdreamer: I've decided to go with `svint16_t svabal[_s16]` and also add a svint16_t svabal[_n_s16] as proposed by @abhishek-kaushik22

rockdreamer:
In the latest force push:
- Neon case is now consistent
- Typo fixes
- svcvt_* is now svcvtzn_*
- svaba_s*[_*] is now just svabal[_s*] and has gained a _n variant
- updated vfdot_lane_f32_f16, vfdotq_laneq_f32_f16, vfdot_laneq_f32_f16, vfdotq_lane_f32_f16 to be similar to their bf16 counterparts.

main/acle.md
```diff
@@ -14581,6 +14778,36 @@ non-overloaded names to indicate which vector argument is a vector register pair
...
+
+#### LUTI6
+
+Lookup table read with 6-bit indices (16-bit).
+
+Use of this intrinsic if `svcntb() * 8 < 512` results in undefined behaviour.
+
+```c
+  // Variants are also available for: _u16, _f16 and _bf16.
+  svint16x4_t svluti6_lane_s16_x4[_s16_x2](svint16x2_t table, svuint8x2_t indices, uint64_t imm_idx);
```
Lukacma:
```suggestion
  svint16x4_t svluti6_lane_s16_x4[_s16_x2](svint16x2_t table, svuint8x2_t indices, uint64_t imm_idx) __arm_streaming;
```

main/acle.md
```diff
@@ -13648,6 +13651,7 @@ Multi-vector saturating rounding shift right narrow and interleave
 Multi-vector saturating rounding shift right unsigned narrow and interleave
 
 ``` c
+  // Variant for _u8[_u16_x2] is available if __ARM_FEATURE_SVE2p3 || __ARM_FEATURE_SME2p3.
```
amilendra: Shouldn't this be `_u8[_s16_x2]`?

rockdreamer: Great catch Amilendra, will fix asap

main/acle.md
```diff
@@ -13857,6 +13953,107 @@ Scalar index of first/last true predicate element (predicated).
...
+  svuint16_t svqshrun[_n]_u16[_s32_x2](svint32x2_t zn, uint64_t imm);
+  ```
+
+#### SUBP
+
+Subtract pairwise.
+
+``` c
+  // Variants are also available for _s16, _s32, _s64, _u8, _u16, _u32 and _u64.
+  svint8_t svsubp[_s8](svbool_t pg, svint8_t zn, svint8_t zm);
```
CarolineConcatto: Should we have here : _x, _z and _m?
See svadd:
https://developer.arm.com/architectures/instruction-sets/intrinsics/#f:@navigationhierarchiesreturnbasetype=[%5Buint,int%5D]&q=svadd

main/acle.md
```diff
@@ -13996,6 +13999,18 @@ Floating-point narrowing convert to interleaved integer, rounding toward zero.
...
+#### SABAL, UABAL
+
+Two-way absolute difference sum and accumulate long.
+
+``` c
+  // Variants are also available for
+  //       _s32, _s64,
+  // _u16, _u32, _u64.
+  svint16_t svabal[_s16](svint16_t zda, svint8_t zn, svint8_t zm);
+  svint16_t svabal[_n_s16](svint16_t zda, svint8_t zn, int8_t zm);
```
amilendra, line 14022: `svint8_t`?

rockdreamer: Arithmetic operations use the disambiguator `_n` when the final operand is a scalar rather than a vector.

main/acle.md
rockdreamer: I completely missed this one. Thanks for catching!

neon_intrinsics/advsimd.md
```diff
@@ -6213,3 +6218,38 @@ The intrinsics in this section are guarded by the macro ``__ARM_NEON``.
...
+|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------|--------------------------------|-------------------|---------------------------|
+| <code>float32x2_t <a href="https://developer.arm.com/architectures/instruction-sets/intrinsics/vfdot_f32_f16" target="_blank">vfdot_f32_f16</a>(<br>&nbsp;&nbsp;&nbsp;&nbsp; float32x2_t r,<br>&nbsp;&nbsp;&nbsp;&nbsp; float16x4_t a,<br>&nbsp;&nbsp;&nbsp;&nbsp; float16x4_t b)</code>                                                           | `r -> Vd.2S`<br>`a -> Vn.4H`<br>`b -> Vm.4H`                     | `FDOT Vd.2S,Vn.4H,Vm.4H`       | `Vd.2S -> result` | `A64`                     |
```
Lukacma: So for better or worse we went with vdot as a name for floating point dot intrinsics (see line 6171). Its best to keep the consistency here.

main/acle.md
```diff
@@ -13857,6 +13953,109 @@ Scalar index of first/last true predicate element (predicated).
...
+  svuint16_t svqshrun[_n]_u16[_s32_x2](svint32x2_t zn, uint64_t imm);
+  ```
+
+#### SUBP
+
+Subtract pairwise.
+
+``` c
+  // Variants are also available for _s16, _s32, _s64, _u8, _u16, _u32 and _u64.
+  svint8_t svsubp[_s8]_m (svbool_t pg, svint8_t zdn, svint8_t zm);
```
CarolineConcatto, line 14061: Just for future reference, I don thins we need a _n version like we have for svsub:
https://developer.arm.com/architectures/instruction-sets/intrinsics/#f:@navigationhierarchiesreturnbasetype=[uint]&q=svsub[
Becausefor SUBP it is subtracting pairs of adjacent  elements WITHIN each source vector, and if we add the _n version it would not make sense.

main/acle.md
```diff
@@ -9950,11 +10026,27 @@ Lookup table read with 4-bit indices.
...
+
+#### LUTI6
+
+Lookup table read with 6-bit indices (8-bit).
+
+Use of this intrinsic if `svcntb() * 8 < 256` results in undefined behaviour.
+
+```c
+  // Variant is  also available for: _u8 _mf8
+  svint8_t svluti6[_s8](svint8x2_t table, svuint8_t indices);
```
CarolineConcatto: Should this be svluti6[_s8_x2]

main/acle.md
```diff
@@ -14648,6 +14687,36 @@ non-overloaded names to indicate which vector argument is a vector register pair
...
+
+#### LUTI6
+
+Lookup table read with 6-bit indices (16-bit).
+
+Use of this intrinsic if `svcntb() * 8 < 512` results in undefined behaviour.
+
+```c
+  // Variants are also available for: _u16, _f16 and _bf16.
+  svint16x4_t svluti6_lane_s16_x4[_s16_x2](svint16x2_t table, svuint8x2_t indices, uint64_t imm_idx);
```
ezrasitorus: Can the [_s16_x2] be dropped also? It seems like there's enough information in _s16_x4 and it's still different from the SVE svluti6_lane[_s16_x2] above.

main/acle.md
rockdreamer: Thanks for catching, will fix

main/acle.md
rockdreamer: Yes, similarly to `svint16_t svluti4_lane[_s16_x2](svint16x2_t table,....`, makes sense.

neon_intrinsics/advsimd.md
rockdreamer: Agreed, thanks for catching

main/acle.md
rockdreamer: Sounds reasonable to me. @Lukacma sounds good?

main/acle.md
rockdreamer: Agreed. Will remove.

main/acle.md
Lukacma: Yeah sounds good. It  should be optional as well

rockdreamer: I think Ezra is proposing not having the optional `[_s16_x2] `at all, 
`svint16x4_t svluti6_lane_s16_x4(svint16x2_t table, svuint8x2_t indices, uint64_t imm_idx);`
This would be similar to how we do   `svint8x2_t svluti2_lane_zt_s8_x2(uint64_t zt, svuint8_t zn,`.

There's also:
- `svint16_t svluti6_lane[_s16_x2](svint16x2_t table, svuint8_t indices, uint64_t imm_idx);`
- `svint16_t svluti4_lane[_s16](svint16_t table, svuint8_t indices, uint64_t imm_idx);`
- `svint16_t svluti4_lane[_s16_x2](svint16x2_t table, svuint8_t indices, uint64_t imm_idx);`

I'm considering this alternative as it's not exactly a lane extraction but a table extraction:
`svint16x4_t svluti6_table[_s16_x2](svint16x2_t table, svuint8x2_t indices, uint64_t imm_idx);`

Thoughts?

Lukacma: So this is the definition of _lane suffix from the ACLE:
```
    *   a `_lane` suffix if they take one multi-vector operand and one
        indexed vector operand with an immediate to specify the indexed
        elements
```
Which fits the form here. If you want to add _table suffix you would need to describe, what it represent and I am not sure it is necessary here. 

Why can't it be   `svint16x4_t svluti6_lane[_s16_x4](svint16x2_t table, svuint8x2_t indices, uint64_t imm_idx);` ?

ezrasitorus: To clarify, I did think of `svint16x4_t svluti6_lane[_s16_x4](svint16x2_t table, svuint8x2_t indices, uint64_t imm_idx);` Perhaps with svluti6_lane, the names are a bit confusing since:

- `svint16_t svluti6_lane[_s16_x2](svint16x2_t table, svuint8_t indices, uint64_t imm_idx)` the suffix refers to the first argument
- `svint16x4_t svluti6_lane[_s16_x4](svint16x2_t table, svuint8x2_t indices, uint64_t imm_idx)` the suffix refers to the return type

rockdreamer: I think keeping the `svint16x4_t svluti6_lane_s16_x4[_s16_x2](svint16x2_t table, svuint8x2_t indices, uint64_t imm_idx)` is correct:
- we want to express the output type because it's different from the input type and inferring it from the inputs complicates the logic quite a bit.
- this intrinsic is quite specific in how it works so I'd keep it as distinct as possible
- I'd keep the optional [_s16_x2] even though it's redundant for uniformity with other intrinsics that contain optional suffixes for their input types.

jthackray (approved):
LGTM

tools/intrinsic_db/advsimd.csv
```diff
@@ -4834,3 +4834,18 @@ float32x4_t vmlallttq_laneq_f32_mf8_fpm(float32x4_t vd, mfloat8x16_t vn, mfloat8
 <SECTION>	Matrix multiplication intrinsics from Armv9.6-A
 float16x8_t vmmlaq_f16_mf8_fpm(float16x8_t r, mfloat8x16_t a, mfloat8x16_t b, fpm_t fpm)	r -> Vd.8H;a -> Vn.16B;b -> Vm.16B	FMMLA Vd.8H, Vn.16B, Vm.16B	Vd.8H -> result	A64
 float32x4_t vmmlaq_f32_mf8_fpm(float32x4_t r, mfloat8x16_t a, mfloat8x16_t b, fpm_t fpm)	r -> Vd.4S;a -> Vn.16B;b -> Vm.16B	FMMLA Vd.4S, Vn.16B, Vm.16B	Vd.4S -> result	A64
+
+<SECTION>	Half-precision dot product to single-precision instructions from Armv9.7-A. Requires the +f16f32dot architecture extension.
+float32x2_t vdot_f32_f16(float32x2_t r, float16x4_t a, float16x4_t b)	r -> Vd.2S;a -> Vn.4H;b -> Vm.4H	FDOT Vd.2S,Vn.4H,Vm.4H	Vd.2S -> result	A64
+float32x4_t vdotq_f32_f16(float32x4_t r, float16x8_t a, float16x8_t b)	r -> Vd.4S;a -> Vn.8H;b -> Vm.8H	FDOT Vd.4S,Vn.8H,Vm.8H	Vd.4S -> result	A64
+
+float32x2_t vdot_lane_f32_f16(float32x2_t r, float16x4_t a, float16x4_t b, __builtin_constant_p(lane))	r -> Vd.2S;a -> Vn.4H;b -> Vm.4H;0 <= lane <= 1	FDOT Vd.2S,Vn.4H,Vm.2H[lane]	Vd.2S -> result	A64
```
Lukacma, line 4842: I am wondering if we should keep non laneq variants. The instruction itself supports only indexing on 128 bit vectors

rockdreamer: The instructions being mapped:
* [FDOT (half-precision to single-precision, by element)](https://developer.arm.com/documentation/ddi0602/2026-03/SIMD-FP-Instructions/FDOT--half-precision-to-single-precision--by-element---Half-precision-dot-product-to-single-precision--vector--by-element--?lang=en)
* [FDOT (half-precision to single-precision, vector)](https://developer.arm.com/documentation/ddi0602/2026-03/SIMD-FP-Instructions/FDOT--half-precision-to-single-precision--vector---Half-precision-dot-product-to-single-precision--vector--?lang=en)

are similar to

* [BFDOT (by element)](https://developer.arm.com/documentation/ddi0602/2026-03/SIMD-FP-Instructions/BFDOT--by-element---BFloat16-dot-product-to-single-precision--vector--by-element--?lang=en)
* [BFDOT (vector)](https://developer.arm.com/documentation/ddi0602/2026-03/SIMD-FP-Instructions/BFDOT--vector---BFloat16-dot-product-to-single-precision--vector--?lang=en)

So I went with what's used for vbfdot_lane_f32 et al (see line 4754)

rockdreamer: BTW, I'll readily admit my eyes blank out every time I try to decode these names. ISTM it's a matter of allowing smaller inputs and limiting the values of lanes?

Lukacma: > So I went with what's used for vbfdot_lane_f32 et al (see line 4754)

I see. If that's how it was done there, then we should keep it consistent

> BTW, I'll readily admit my eyes blank out every time I try to decode these names. ISTM it's a matter of allowing smaller inputs and limiting the values of lanes?

I am not sure I understand the question ?

rockdreamer: It was me trying to understand why the existing intrinsics work the way they do. I'll mark this as resolved.

main/acle.md
```diff
@@ -13863,6 +13959,108 @@ Scalar index of first/last true predicate element (predicated).
...
+
+#### SCVTF, SCVTFLT, UCVTF, UCVTFLT
+
+Integer convert to floating-point (top and bottom).
+
+``` c
+  // Variants are also available for
+  //            _f32[_s16], _f64[_s32],
+  // _f16[_u8], _f32[_u16], _f64[_u32].
+  svfloat16_t svcvtt_f16[_s8](svint8_t zn);
```
Lukacma, line 14033: I think return type doesn't need to be part of the overloaded names for these conversion intrinsics.

main/acle.md
```diff
@@ -13863,6 +13959,108 @@ Scalar index of first/last true predicate element (predicated).
...
+
+#### FCVTZSN, FCVTZUN
+
+Floating-point narrowing convert to interleaved integer, rounding toward zero.
+
+``` c
+  // Variants are also available for
+  //               _s16[_f32_x2], _s32[_f64_x2],
+  // _u8[_f16_x2], _u16[_f32_x2], _u32[_f64_x2].
+  svint8_t svcvtzn_s8[_f16_x2](svfloat16x2_t zn);
```
Lukacma: I think return type doesn't need to be part of the overloaded names for these conversion intrinsics.

main/acle.md
rockdreamer: I agree with you in that the return type can be inferred from the input type and is strictly optional in this case.

However, all sve conversion functions start with the main returned data type after the stem.
A search in the [instrinsics db](https://developer.arm.com/architectures/instruction-sets/intrinsics/#f:@navigationhierarchiesreturnbasetype=[%5Buint%5D,uint]&q=svcvt) and in acle.md shows that this is an unwritten but strictly adhered to rule.
There's inconsistency in that sometimes you get the return vector size in weird places (like `svfloat16x2_t svcvtl1_f16[_mf8]_x2_fpm(svmfloat8_t zn, fpm_t fpm)`) or not at all (like `svfloat32x2_t svcvtl_f32[_f16_x2](svfloat16_t zn)`), but the main data type returned is always there.

There's also the possibillty that a version returning bf16 gets added in the future and having the return type after the stem leaves this option open.
Hopefully this explains my preference for keeping this function and svcvtzn_s8 below the way they are.

main/acle.md
rockdreamer: Agreed, but as explained above for svcvtt_f16, it's there for consistency between svcvt instrinsics.

main/acle.md
Lukacma: Yes. That is valid point and I agree we should try to be consistent here.

main/acle.md
```diff
@@ -13864,6 +13960,108 @@ Scalar index of first/last true predicate element (predicated).
...
```
Lukacma: For consistency, shouldn't this be just scvtn ?

rockdreamer: FCVTN uses this [rounding function](https://developer.arm.com/documentation/ddi0487/mb/-Part-J-Architectural-Pseudocode/-Chapter-J1-A-profile-Architecture-Pseudocode/-J1-4-Shared-pseudocode/-J1-4-286-FP8Round?lang=en#asl_func_fp8round_5) that depends on the FPCR rounding mode.
Whereas  FCVTZSN always rounds towards zero.

Lukacma: So we have other [intrinsics](https://developer.arm.com/architectures/instruction-sets/intrinsics/#f:@navigationhierarchiessimdisa=[SME%20and%20SME2,SVE2,SVE]&q=FCVTZS) which use rounding toward zero instructions and we don't specify that in the intrinsics name.

rockdreamer: I think you've uncovered a different inconsistency.
You're right that the svcvt_ conversions narrow towards zero (TIL!).
However:
- `svmfloat8_t svcvt_mf8[_f16_x2]_fpm(svfloat16x2_t zn, fpm_t fpm)` uses BFCVT, FCVT which has a complex rounding mode
- `svbfloat16_t svcvt_bf16[_f32_x2](svfloat32x2_t zn)` use FCVT, BFCVT which have complex rounding

There is little to no indication that these behave differently. I think we should improve the documentation for these.

Existing [svcvtn_ instrinsics](https://developer.arm.com/architectures/instruction-sets/intrinsics/#f:@navigationhierarchiessimdisa=[%5BSME%20and%20SME2,SVE2,SVE%5D]&q=svcvtn_) have complex rounding.
So do:
- `svbfloat16_t svcvtn_bf16[_f32_x2](svfloat32x2_t zn)`
- `svmfloat8_t svcvtn_mf8[_f16_x2]_fpm(svfloat16x2_t zn, fpm_t fpm)`

However they all deal with floating point conversions. I guess the fact that `svint8_t svcvtzn_s8[_f16_x2](svfloat16x2_t zn)` does a conversion to integer means we can skip the z in this case.

Lukacma: Yeah I think that's the assumption. Let's hope they never add conversion to integers with different rounding behaviour 🤞

tools/intrinsic_db/advsimd.csv
Lukacma, line 4842: Do we need the _f16 suffix for these intrinsics? We don't have second suffix for integer variants.

rockdreamer: Yep, for consistency with `float32x2_t vdot_f32_mf8_fpm`
Unlike SVE intrinsics, AdvSIMD intrinsics do not make use of overloading so we need to make the type they operate more explicit.
`float32x2_t vbfdot_f32` does not need the `_bf16` suffix because bfdot is specific to the bf16 type.
I believe the other dot functions also have a prefix that makes the target type clear.
Given we're using the common vdot prefix, we must use a postfix for the types.

main/acle.md
```diff
@@ -489,6 +489,22 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * Added [**Alpha**](#current-status-and-anticipated-changes) support
   for Brain 16-bit floating-point vector multiplication intrinsics.
 * Redesigned atomic store with hints intrinsics.
+* Added [**Alpha**](#current-status-and-anticipated-changes)
```
Lukacma, line 492: These changes were not part of ACLE Q1 2026 release

main/acle.md
```diff
@@ -13864,6 +13960,108 @@ Scalar index of first/last true predicate element (predicated).
...
+  ```
+
+#### LUTI6
+
+Lookup table read with 6-bit indices (16-bit).
+
+Use of this intrinsic if `svcntb() * 8 < 512` results in undefined behaviour.
+
+``` c
+  // Variants are also available for _u16_x2 and _f16_x2.
```
jthackray, line 13998:
```suggestion
  // Variants are also available for _u16_x2, _f16_x2 and _bf16_x2.
```
Sorry I didn't spot this sooner!

rockdreamer: Thanks for catching this!

main/acle.md
rockdreamer: I'll post a patch ASAP

PR #430, opened 2026-03-25
Add FEAT_CMH support

ricbal02:
---
Hi all,

This is a PR to propose the addition of new intrinsics required as part of FEAT_CMH.

Quality Level: ALP

---

**Thank you for submitting a pull request!**

If this PR is about a bugfix:

Please use the bugfix label and make sure to go through the checklist below.

If this PR is about a proposal:

We are looking forward to evaluate your proposal, and if possible to
make it part of the Arm C Language Extension (ACLE) specifications.

We would like to encourage you reading through the [contribution
guidelines](https://github.com/ARM-software/acle/blob/main/CONTRIBUTING.md), in particular the section on [submitting
a proposal](https://github.com/ARM-software/acle/blob/main/CONTRIBUTING.md#proposals-for-new-content).

Please use the proposal label.

As for any pull request, please make sure to go through the below
checklist.

Checklist: (mark with ``X`` those which apply)

* [ ] If an issue reporting the bug exists, I have mentioned it in the
      PR (do not bother creating the issue if all you want to do is
      fixing the bug yourself).
* [ ] I have added/updated the `SPDX-FileCopyrightText` lines on top
      of any file I have edited. Format is `SPDX-FileCopyrightText:
      Copyright {year} {entity or name} <{contact informations}>`
      (Please update existing copyright lines if applicable. You can
      specify year ranges with hyphen , as in `2017-2019`, and use
      commas to separate gaps, as in `2018-2020, 2022`).
* [ ] I have updated the `Copyright` section of the sources of the
      specification I have edited (this will show up in the text
      rendered in the PDF and other output format supported). The
      format is the same described in the previous item.
* [ ] I have run the CI scripts (if applicable, as they might be
      tricky to set up on non-*nix machines). The sequence can be
      found in the [contribution
      guidelines](https://github.com/ARM-software/acle/blob/main/CONTRIBUTING.md#continuous-integration). Don't
      worry if you cannot run these scripts on your machine, your
      patch will be automatically checked in the Actions of the pull
      request.
* [ ] I have added an item that describes the changes I have
      introduced in this PR in the section **Changes for next
      release** of the section **Change Control**/**Document history**
      of the document. Create **Changes for next release** if it does
      not exist. Notice that changes that are not modifying the
      content and rendering of the specifications (both HTML and PDF)
      do not need to be listed.
* [ ] When modifying content and/or its rendering, I have checked the
      correctness of the result in the PDF output (please refer to the
      instructions on [how to build the PDFs
[...]

PR #432, opened 2026-04-16
Redesign atomic store with hints intrinsics

Lukacma:
This proposal is about redesigning atomic store with hint intrinsics, to make them more easily extensible. 

      fixing the bug yourself).
* [x ] I have added/updated the `SPDX-FileCopyrightText` lines on top
* [ x] I have updated the `Copyright` section of the sources of the
      found in the [contribution
      request.
* [x ] I have added an item that describes the changes I have
      do not need to be listed.
* [ x] When modifying content and/or its rendering, I have checked the
* [ x] The variable `draftversion` is set to `true` in the YAML header

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -487,6 +487,7 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
 * Removed all references to Transactional Memory Extension (TME).
 * Added [**Alpha**](#current-status-and-anticipated-changes) support
   for Brain 16-bit floating-point vector multiplication intrinsics.
+* Redesigned atomic store with hints intrinsics.
 
 ### References
 
--- a/main/acle.md
+++ b/main/acle.md
@@ -11,7 +11,7 @@ toc: true
 ---
 
 <!--
-SPDX-FileCopyrightText: Copyright 2011-2025 Arm Limited and/or its affiliates <open-source-office@arm.com>
+SPDX-FileCopyrightText: Copyright 2011-2026 Arm Limited and/or its affiliates <open-source-office@arm.com>
 SPDX-FileCopyrightText: Copyright 2022 Google LLC.
 CC-BY-SA-4.0 AND Apache-Patent-License
 See LICENSE.md file for details
@@ -114,7 +114,7 @@ about Arm’s trademarks.
 
 ## Copyright
 
-* Copyright 2011-2024 Arm Limited and/or its affiliates <open-source-office@arm.com>.
+* Copyright 2011-2026 Arm Limited and/or its affiliates <open-source-office@arm.com>.
 * Copyright 2022 Google LLC.
 
 ## About this document
--- a/main/acle.md
+++ b/main/acle.md
@@ -4960,7 +4960,7 @@ The hint is a suggestion to the compiler and maps directly to a
 specific hint instruction variant in the ISA. The compiler may use this hint
 when selecting code sequences, but it is not required to emit a specific
 hint instruction or a specific instruction sequence. This intrinsic is
-type generic and supports scalar integer and pointer types of 8, 16, 32, and 64 bits.
+type generic and supports scalar integral and floating-point types of 8, 16, 32, and 64 bits.
 
 To access this intrinsic, `<arm_acle.h>` should be included.
 
--- a/main/acle.md
+++ b/main/acle.md
@@ -4952,33 +4952,38 @@ stored to memory is modified by replacing the low 32 bits of
 `value.val[0]` with the contents of the `ACCDATA_EL1` system register.
 The returned value is the same as for `__arm_st64bv`.
 
-## Atomic store with PCDPHINT intrinsics
+## Atomic store with hints intrinsics
 
-This intrinsic provides an atomic store, which will
-make use of the `STSHH` hint instruction immediately followed by the
-associated store instruction. This intrinsic is type generic and 
-supports scalar types from 8-64 bits and is available when
-`__ARM_FEATURE_PCDPHINT` is defined.
+This intrinsic provides an atomic store together with a hint value.
+The hint is a suggestion to the compiler and maps directly to a
+specific hint instruction variant in the ISA. The compiler may use this hint
+when selecting code sequences, but it is not required to emit a specific
+hint instruction or a specific instruction sequence. This intrinsic is
+type generic and supports scalar integer and pointer types of 8, 16, 32, and 64 bits.
 
 To access this intrinsic, `<arm_acle.h>` should be included.
 
 ``` c
-  void __arm_atomic_store_with_stshh(type *ptr,
-                                     type data,
-                                     int memory_order,
-                                     int ret); /* Retention Policy */
+  void __arm_atomic_store_with_hint(type *ptr,
+                                    type data,
+                                    int memory_order,
+                                    int hint);
 ```
 
-The first argument in this intrinsic is a pointer `ptr` which is the location to store to.
-The second argument `data` is the data which is to be stored.
-The third argument `mem` can be one of 3 memory ordering variables supported by atomic_store:
-__ATOMIC_RELAXED, __ATOMIC_SEQ_CST, and __ATOMIC_RELEASE.
-The fourth argument can contain the following values:
+The first argument `ptr` is the location to store to. The second
+argument `data` is the value to be stored. The third argument
+`memory_order` can be one of the memory ordering values supported by
+`atomic_store`: `__ATOMIC_RELAXED`, `__ATOMIC_SEQ_CST`, and
+`__ATOMIC_RELEASE`.
 
-| **Retention Policy** | **Value** | **Summary**                                                                       |
-| -------------------- | --------- | --------------------------------------------------------------------------------- |
-| KEEP                 | 0         | Signals to retain the updated location in the local cache of the updating PE.     |
-| STRM                 | 1         | Signals to not retain the updated location in the local cache of the updating PE. |
+The fourth argument `hint` selects the requested hint. The set of valid
+hint values depends on the architectural features supported by the
+target. The following hint values are defined:
+
+| **Hint**         | **Value** | **Feature**                | **Summary**                                                                       |
+| ---------------- | --------- | -------------------------- | --------------------------------------------------------------------------------- |
+| HINT_STSHH_KEEP  | 0         | `__ARM_FEATURE_PCDPHINT`   | Requests retention of the updated location in the local cache of the updating PE. |
+| HINT_STSHH_STRM  | 1         | `__ARM_FEATURE_PCDPHINT`   | Requests that the updated location not be retained in the local cache of the updating PE. |
 
 # Custom Datapath Extension
 


main/acle.md
```diff
@@ -4952,33 +4953,38 @@ stored to memory is modified by replacing the low 32 bits of
...
+argument `data` is the value to be stored. The third argument
+`memory_order` can be one of the memory ordering values supported by
+`atomic_store`: `__ATOMIC_RELAXED`, `__ATOMIC_SEQ_CST`, and
+`__ATOMIC_RELEASE`.
 
-| **Retention Policy** | **Value** | **Summary**                                                                       |
-| -------------------- | --------- | --------------------------------------------------------------------------------- |
-| KEEP                 | 0         | Signals to retain the updated location in the local cache of the updating PE.     |
-| STRM                 | 1         | Signals to not retain the updated location in the local cache of the updating PE. |
+The fourth argument `hint` selects the requested hint. The set of valid
```
CarolineConcatto, line 4980: What will be the behaviour if hint is not valid?

main/acle.md
```diff
@@ -4952,33 +4953,38 @@ stored to memory is modified by replacing the low 32 bits of
...
 
-The first argument in this intrinsic is a pointer `ptr` which is the location to store to.
-The second argument `data` is the data which is to be stored.
-The third argument `mem` can be one of 3 memory ordering variables supported by atomic_store:
-__ATOMIC_RELAXED, __ATOMIC_SEQ_CST, and __ATOMIC_RELEASE.
-The fourth argument can contain the following values:
+The first argument `ptr` is the location to store to. The second
+argument `data` is the value to be stored. The third argument
+`memory_order` can be one of the memory ordering values supported by
+`atomic_store`: `__ATOMIC_RELAXED`, `__ATOMIC_SEQ_CST`, and
```
CarolineConcatto, line 4977: Do we need to list all the possible memory order :
```
``__ATOMIC_RELAXED``, ``__ATOMIC_CONSUME``, ``__ATOMIC_ACQUIRE``,
``__ATOMIC_RELEASE``, ``__ATOMIC_ACQ_REL``, and ``__ATOMIC_SEQ_CST`` are
provided, with values corresponding to the enumerators of C11's
```
Or the ones you are defining is enough? I am asking because this will be a generic intrinsic now. So I am wondering if all the hint will only need these 3 macros.

main/acle.md
```diff
@@ -4952,33 +4953,38 @@ stored to memory is modified by replacing the low 32 bits of
...
-| KEEP                 | 0         | Signals to retain the updated location in the local cache of the updating PE.     |
-| STRM                 | 1         | Signals to not retain the updated location in the local cache of the updating PE. |
+The fourth argument `hint` selects the requested hint. The set of valid
+hint values depends on the architectural features supported by the
+target. The following hint values are defined:
+
+| **Hint**         | **Value** | **Feature**                | **Summary**                                                                       |
+| ---------------- | --------- | -------------------------- | --------------------------------------------------------------------------------- |
+| HINT_STSHH_KEEP  | 0         | `__ARM_FEATURE_PCDPHINT`   | Requests retention of the updated location in the local cache of the updating PE. |
+| HINT_STSHH_STRM  | 1         | `__ARM_FEATURE_PCDPHINT`   | Requests that the updated location not be retained in the local cache of the updating PE. |
```
rockdreamer, line 4987: We don't have a crystal ball into the future so I wonder if we might find ourselves into a situation where more than one hint might be possible.
What if we made the interface composable with 0 being HINT_NONE, HINT_STSHH_KEEP 2, HINT_STSHH_STRM 4 etc.
Pros:
- composability

Cons:
- limits the hint space to 63 values

Depends on who you ask:
- This can gracefully ignore future hints the compiler doesn't know about.

main/acle.md
Lukacma: __ATOMIC_CONSUME  and __ATOMIC_ACQUIRE is only for loads. __ATOMIC_ACQ_REL doesn't fit pure store I think and rest are mentioned in spec.

main/acle.md
CarolineConcatto: Can we just not write one after another of the intrinsic above? Like this;
 __arm_atomic_store_with_hint(*ptr,   data, memory_order,   hint1);
 __arm_atomic_store_with_hint( *ptr,  data,  memory_order,   hint2);

Lukacma: So we had discussion about this internally, and there was an argument that we do not need to do that. It seems that architects plan to implement chaining of hints as separate encoding. For example, [SHUH](https://developer.arm.com/documentation/ddi0602/2026-03/Base-Instructions/SHUH--Shared-update-hint-) hint has an option to additionally provide STCPH hint.

Lukacma: I was also looking at the spec and I am not sure chaining is even possible for current hints. Looking at [SHUH](https://developer.arm.com/documentation/ddi0602/2026-03/Base-Instructions/SHUH--Shared-update-hint-) spec:
This instruction signals to the memory system that if the next instruction in program order generates an Explicit Memory Effect,So if we chained SHUH followed by another hint than it seems to me that the SHUH would be ignored 

The other hints have similiar wording so I don't think that this is smth we need to futureproof for.

rockdreamer: Cool, thanks for the clarification. 👍

main/acle.md
Lukacma: I guess that is undefined behaviour and will be implementation based. For LLVM we will throw error and reject.

CarolineConcatto: Should it throw an error or just ignore and still do the atomic store?

Lukacma: I would throw error. This seems like a user mistake and we shouldn't just let it silently pass imo.

main/acle.md
```diff
@@ -4952,33 +4953,38 @@ stored to memory is modified by replacing the low 32 bits of
...
+`memory_order` can be one of the memory ordering values supported by
+`atomic_store`: `__ATOMIC_RELAXED`, `__ATOMIC_SEQ_CST`, and
+`__ATOMIC_RELEASE`.
 
-| **Retention Policy** | **Value** | **Summary**                                                                       |
-| -------------------- | --------- | --------------------------------------------------------------------------------- |
-| KEEP                 | 0         | Signals to retain the updated location in the local cache of the updating PE.     |
-| STRM                 | 1         | Signals to not retain the updated location in the local cache of the updating PE. |
+The fourth argument `hint` selects the requested hint. The set of valid
+hint values depends on the architectural features supported by the
```
sallyarmneale, line 4981: Is it Hint or hint?

CarolineConcatto: What is the difference between Hint and hint besides the first letter being in capital or lower? Does it have different meaning for @sallyarmneale ?

CarolineConcatto (approved):

main/acle.md
ricbal02: I agree with throwing an error, the generic versions of these should be used when not using a hint

ricbal02:
LGTM

main/acle.md
paulwalker-arm: Given the point of this change is to make the builtin more generic, it doesn't make sense to then add barriers to future uses. We've already said a toolchain is free to ignore the hint and that's effectively what an older toolchain should do if given a hint that only newer toolchains support.  If we want to help users spot potential typos then we can add an "unknown hint will be ignored" diagnostic.

main/acle.md
```diff
@@ -4952,33 +4953,38 @@ stored to memory is modified by replacing the low 32 bits of
...
-make use of the `STSHH` hint instruction immediately followed by the
-associated store instruction. This intrinsic is type generic and 
-supports scalar types from 8-64 bits and is available when
-`__ARM_FEATURE_PCDPHINT` is defined.
+This intrinsic provides an atomic store together with a hint value.
+The hint is a suggestion to the compiler and maps directly to a
+specific hint instruction variant in the ISA. The compiler may use this hint
+when selecting code sequences, but it is not required to emit a specific
+hint instruction or a specific instruction sequence. This intrinsic is
+type generic and supports scalar integer and pointer types of 8, 16, 32, and 64 bits.
```
rockdreamer: Should we remove integer here (or add a specific list of types like char and floating point)?

rockdreamer (commented):
LGTM otherwise

main/acle.md
ricbal02: Definitely yes on removing integer as Float and Double are supported. I suspect a list is probably unnecessary

main/acle.md
ricbal02: So for the atomic_store intrinsics here, it would be fine to have them work without the hint, however, for the atomic_fetch intrinsics added in #430, the assembly output is limited to the LSE variant as the hint does not work for other variants, so outline atomics will not be used. In that case we really don't want this to be seen as an alternative to using the non-hint atomics

main/acle.md
Lukacma: Okay finished digging and there are definitely changes to be made. The first question we need to clarify is whether we are implementing` void __atomic_store_n (T *obj, T val, memory_order m)  `or void `__atomic_store (T* object, T* new_value, memory_order m)` , both  or something custom in terms of base atomics behaviour. If first, then we support only integral types of sizes 8,16,32,64,**128** . If 2nd we need to change a signature of the function to use pointer for value and we need to support **all types** including  weird types like neon vector types (Looks like scalable types don't work, which makes sense) . If both we need new intrinsics. If we want smth custom then we need to discuss what types we want to support and why are we excluding the others. Here is my [godbolt snippet ](https://godbolt.org/z/qGa4qE9Tz)showing my experiments on this and it is consistent with documentation [here](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html) and [here](https://gcc.gnu.org/wiki/Atomic/GCCMM/LIbrary)

ricbal02: FWIW I have already implemented this intrinsic is GCC16, which was just released, using the ACLE spec I wrote for PCDPHINT, so any major change like this is probably best avoided. For that All scalars including FP between 8-64 bits were included

ricbal02: These hint intrinsics have more niche use cases than the generic atomics, therefore we don't necessarily need to match 1:1

Lukacma: > FWIW I have already implemented this intrinsic is GCC16, which was just released, using the ACLE spec I wrote for PCDPHINT, so any major change like this is probably best avoided. For that All scalars including FP between 8-64 bits were included

I am not sure this is relevant. If I understand correctly you implemented the old intrinsic, which is going to be removed anyway so its semantics don't really matter ? Or am I wrong ? 

> These hint intrinsics have more niche use cases than the generic atomics, therefore we don't necessarily need to match 1:1

True, but I think we shouldn't be arbitrary in the semantics of these intrinsics and probably stick as close as possible to the sematics of the base atomics as possible, unless we have good reason to deviate. Otherwise user will be confused why our intrinsics don't work in situations where normal ones would. This is specially important here as these intrinsics should really be just base atomic store + hint optionally attached in front of i it in assembly.

ricbal02: I've already written a change from the old intrinsic to the new, from an implementation point of view this is a minor change. I can see the sense in that argument, well the current intrinsics are like atomic_store_n, however I have added support for FP, there is no support for 128 bit at current, I think that is something that could be added later. I could remove support for FP to make it exactly like atomic_store_n but I don't see a benefit in that.

Lukacma: We have been talking about this with @paulwalker-arm  and I don't think we are strongly against adding FP support. However if we do that we are completely diverging from spec of  GCC's atomic_store and atomic_store_n intrinsics, which is not an issue as LLVM has support for floating point atomics. But going in that direction, we might as well make the intrinsic more general. As this intrinsic will have different behaviour than GCC atomics, it would be useful to allow users to use it for floating point atomics even without a hint. What do you think ? 

As for 128 bits integers, it seems like both compilers are able to support it https://godbolt.org/z/ahxr1x4sd , so I think it wouldn't be too difficult to implement and it would make the intrinsic more complete imo, but if you are strongly against that we can add it later as you suggested.

ricbal02: I'm not completely against having them be usable without a hint, I think it's highly unlikely anyone will end up using it that way. In theory the best course of action would be to have atomic_store_n allow FP in GCC, however that isn't going to happen any time soon. If you feel strongly about then I won't stop you adding a no hint option. The work required in implementation for 128 bit atomic is not necessarily straight forward, I think it is best left alone for now

Lukacma: I’m a bit confused here. If the concern is that FP atomics won’t be useful without the hint, what makes the hinted version different? I’m not seeing what is specific to hints that would make FP atomics more useful in that case ?

ricbal02: FP atomics are useful, but atomic_store already exists in GCC which allows the use of FP, so it's unlikely that someone would use this new one instead

Lukacma: Ah yes ! That is what I missed. Thanks for clarifying. I will add floating point types to the spec and leave 128 bit variant be for now.

rockdreamer (approved):
LGTM

kmclaughlin-arm (approved):

PR #435, opened 2026-05-15
Fix specification of FP16 scalar conversion to/from fixed point

momchil-velikov:
Related to https://github.com/llvm/llvm-project/issues/197083
---
name: Pull request

---

If this PR is about a bugfix:

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.


CarolineConcatto (commented):
I understand that GCC and LLVM use different instructions, which is inconsistent, but changing this would also change the behaviour of the intrinsic.Do we want to do that after seven years?

momchil-velikov:
> I understand that GCC and LLVM use different instructions, which is inconsistent, but changing this would also change the behaviour of the intrinsic.Do we want to do that after seven years?

Taking as an example [`vcvth_n_f16_s32`](https://developer.arm.com/architectures/instruction-sets/intrinsics/#q=vcvth_n_f16_s32) 

The specification says:

> Signed fixed-point Convert to Floating-point (vector). This instruction converts each element in a vector from fixed-point to floating-point using the rounding mode that is specified by the FPCR, and writes the result to the SIMD&FP destination register.

Let's ignore for a moment that spec says "vector" and focus on the important part, which is the numerical behaviour of the intrinsic - it converts using rounding mode specified by FPCR.

However, the mentioned instruction takes its source operand from a 16-bit register.  There's no guarantee that the discarded bits (since the original argument is 32-bit) do not affect the final result for all rounding modes. In that sense, the specification either self-contradicting, unimplementable, or implicitly assumes the compiler does some **non-trivial** conversion of a 32-bit value to 16 bits, in which case it'd be incomplete.

Therefore it's a bug in the specification that needs to be fixed, and here's my proposal how to fix it, which is consistent with the original intent of using FPCR rounding mode, improves precision, is at least  as efficient as the previous approach, is consistent with GCC implementation, and is simple in the sense it just changes the instruction (as opposed to specifying an eleborate conversion sequence that ensures final result using the current instruction is rounded as specified by FPCR).

Now the question is, does the change in behaviour negatively affects users?  There are a few factors that reduce the exposure area:
*  FP16 not widely used (compared to `float` and `double`)
*  it's used mostly in graphics and AI, where the exact bit precision is not terribly important
*  a large number of the potentially affected users are using GCC anyway

That is to say, while there could be negatively affected users, it's unlikely.
In my opinion the benefits of correct specification (and further down the line, implementation), and compatibility with GCC far outweigh the potential drawbacks.

momchil-velikov:
Another option would be to evolve the specification over a period of time:
 * deprecate the current behavior, with a remark in the specification
 * issue a warning in Clang when the affected intrinsics are used that the behavior is about to change
 * (optionally) provide compatibility intrinsics retaining the current behavior (which would affect GCC as well)
 * at the end of the deprecation period (1 year, 2 years?)  update the specification and Clang

rockdreamer:
This is clearly ok from the gcc side :)

CarolineConcatto:
So, it looks like we should update this on the llvm side, as the behaviour was incorrect.

CarolineConcatto (approved):

PR #439, opened 2026-07-06
Explicitly forbid inheriting SME keyword attributes

acoplan-arm:
This patch aims to clarify the interaction between SME keyword attributes and a common compiler extension which allows type attributes to be inherited given a pair of subsequent declarations D1, D2 where D1 has type attributes and D2 has none.

The patch just explicitly spells out behaviour which (in my understanding) is:
 - the intent of the existing specification
 - existing clang behaviour
 - intended GCC behaviour (which has regressed in GCC 16, see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=122483).  The hope is that the new wording will help smooth the way for this regression to be fixed in GCC 17.

The SME keyword attributes section says:

> Except where noted otherwise, function types that have an attribute
> are incompatible with function types that do not.

I believe the intent here was for this incompatibility to extend to compiler extensions which currently allow code such as:

```c
void f(void) [[attr1, ..., attrN]];
void f(void);
```

as a convenience for situations where f is likely to be included from a header file, and the user doesn't want to have to write out all of the attributes in a local redeclaration.

E.g. the following C code is currently accepted by both GCC and clang:

```c
void f(void) __attribute__((aarch64_vector_pcs));
void f(void); // accepted
```

but if we switch the order of the declarations, it is (correctly) rejected by both compilers:

```c
void g(void);
void g(void) __attribute__((aarch64_vector_pcs)); // rejected
```

I believe the intent of the ACLE for SME keyword attributes was always to reject code such as:

```cpp
void f(void) __arm_inout("za");
void f(void);
```

i.e. that this convenience feature should _not_ apply to those SME keyword attributes that affect type identity.  The rationale is along the lines that the above is logically equivalent to something like:

```cpp
void f(za_state &);
void f(void);
```

which should be considered ill-formed in the same way that e.g.:

```cpp
int f(int x);
void f(void);
```

is.

### DIFF
--- a/main/acle.md
+++ b/main/acle.md
@@ -510,6 +510,9 @@ Armv8.4-A [[ARMARMv84]](#ARMARMv84). Support is added for the Dot Product intrin
   support for SVE2.3 (FEAT_SVE2p3) and SME2.3 shift right narrow intrinsics.
 * Bumped armv9.6 intrinsics implementation to [**Beta**](#current-status-and-anticipated-changes)
 * Added support for producer-consumer data placement hints.
+* Clarified interaction between SME keyword attributes and a common compiler
+  extension whereby type attributes can be inherited between subsequent
+  duplicate decls.
 
 ### References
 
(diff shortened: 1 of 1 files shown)


rockdreamer (approved):
LGTM

paulwalker-arm (approved):

main/acle.md
```diff
@@ -10743,6 +10746,35 @@ The function type attributes cannot be used with K&R-style
...
+
+```c
+void f(void) __arm_streaming;
+void f(void); // Ill-formed
+
+void g(void) __arm_streaming_compatible;
+void g(void) {} // Ill-formed
+
+void h(void) __arm_streaming __arm_inout("za");
+void h(void); // Ill-formed
```
alicecarlotti, line 10775: Slight nit - did you mean to include __arm_streaming on the second declaration, so that only the __arm_inout differs?

Fwiw, the patch looks good otherwise.

acoplan-arm: Are you saying it would be clearer without the `__arm_streaming` on `h`? I was trying to give an idiomatic example (which typically has both keyword attributes), but I'm happy to change it in a follow-up patch if you think it's more illustrative of the situation without.

I suppose another option is just adding another example with only the `__arm_inout("za")` part differing.

Thanks for the review.

paulwalker-arm: These are just examples, we don't need to specify every combination.

PR #440, opened 2026-07-07
Add missing Neon intrinsics

MartinWehking:
These intrinsics are present in LLVM and GCC, but not documented on the ACLE.
Add them to close the gap.

---
name: Pull request

---

If this PR is about a bugfix:

checklist.

      fixing the bug yourself).
      found in the [contribution
      request.
      do not need to be listed.


neon_intrinsics/advsimd.md
```diff
@@ -129,6 +129,7 @@ for more information about Arm’s trademarks.
 | M     | 21 February 2025  | 2024Q4               |
 | N     | 06 June 2025      | 2025Q2               |
 | O     | 15 May 2026       | 2026Q1               |
+| P     | 07 July 2026      | 2026Q2               |
```
CarolineConcatto: We will not have releases anymore, so maybe we dont need to change this anymore

neon_intrinsics/advsimd.md
```diff
@@ -174,6 +175,7 @@ for more information about Arm’s trademarks.
 
 * Added support for FEAT_F16F32DOT
 * Added support for FEAT_F16F32MM and FEAT_F16MM
+* Added documentation for missing Neon intrinsics
```
CarolineConcatto, line 176: Same with this. Maybe we should change this section to be only:
### Changes since 2026Q1:

neon_intrinsics/advsimd.md
```diff
@@ -3293,6 +3307,8 @@ The intrinsics in this section are guarded by the macro ``__ARM_NEON``.
...
 | <code>poly16x4_t <a href="https://developer.arm.com/architectures/instruction-sets/intrinsics/vmov_n_p16" target="_blank">vmov_n_p16</a>(poly16_t value)</code>                                                                                        | `value -> rn`                        | `DUP Vd.4H,rn`          | `Vd.4H -> result`  | `v7/A32/A64`              |
 | <code>poly16x8_t <a href="https://developer.arm.com/architectures/instruction-sets/intrinsics/vmovq_n_p16" target="_blank">vmovq_n_p16</a>(poly16_t value)</code>                                                                                      | `value -> rn`                        | `DUP Vd.8H,rn`          | `Vd.8H -> result`  | `v7/A32/A64`              |
+| <code>poly64x1_t <a href="https://developer.arm.com/architectures/instruction-sets/intrinsics/vmov_n_p64" target="_blank">vmov_n_p64</a>(poly64_t value)</code>                                                                                        | `value -> rn`                        | `INS Dd.D[0],xn`        | `Vd.1D -> result`  | `A32/A64`                 |
```
CarolineConcatto: Is this correct ' INS Dd.D[0],xn`, it is different from the others vmovq  with poly types.

CarolineConcatto: Should we change this to also report v7?

tools/intrinsic_db/advsimd.csv
```diff
@@ -1967,6 +1969,8 @@ poly8x8_t vmov_n_p8(poly8_t value)	value -> rn	DUP Vd.8B,rn	Vd.8B -> result	v7/A
 poly8x16_t vmovq_n_p8(poly8_t value)	value -> rn	DUP Vd.16B,rn	Vd.16B -> result	v7/A32/A64
 poly16x4_t vmov_n_p16(poly16_t value)	value -> rn	DUP Vd.4H,rn	Vd.4H -> result	v7/A32/A64
 poly16x8_t vmovq_n_p16(poly16_t value)	value -> rn	DUP Vd.8H,rn	Vd.8H -> result	v7/A32/A64
+poly64x1_t vmov_n_p64(poly64_t value)	value -> rn	INS Dd.D[0],xn	Vd.1D -> result	A32/A64
```
CarolineConcatto: This looks strange too.

neon_intrinsics/advsimd.md
MartinWehking: I wish I could give a strong reason why it should be/ shouldn't be v7, but all I can do is base it on patterns:

`vmovq_n_p64` also uses `A32/A64` only.
Most other  `_p64`intrinsics such as vcombine_p64 are also just `A32/A64`.

These intrinsics were also present in LLVM/gcc. Inside the LLVM backend they are currently guarded only by A64 flags, but should be A32 probably as well.

Do you think it should be v7 or do you perhaps know how I could definitely say it is/isn't v7?

MartinWehking: For both GCC and LLVM, the compiler is emitting a `fmov` in assembly.
Fr other intrinsics `fmov` is an INS rather than a DUP:
https://godbolt.org/z/WojWWKrss

vmovq on the other hand emits a "real" DUP in assembly.

Compare vdup_n_p64, which is also emitting an INS only.

tools/intrinsic_db/advsimd.csv
MartinWehking: This is the original input file. The md gets generated from the csv, so I set the value here. The strangeness was carried into the md file, so to speak

neon_intrinsics/advsimd.md
georges-arm: This doesn't make sense as is:
* `INS` here implies a partial (merging) write to the destination register but this intrinsic does not specify a destination to merge into (it wouldn't make sense to do so either).
* `INS` with a `Dd` (not `Vd`) destination doesn't exist in AArch64 at least.

So I think this should be `FMOV` since that is the closest correct interpretation (and that's what compilers currently emit anyway for the GPR->VPR mov).

georges-arm: The primary use case for 64-bit polynomial types would be to use the PMULL instruction, but that is limited to A32/A64 as well, so I think the restriction makes sense here (but it's not really a problem to have it in v7 either so no strong preference).

Same comment about `INS` not making sense here.

MartinWehking: > INS with a Dd (not Vd) destination doesn't exist in AArch64 at least.

`vdup_n_p64` is also wrong then: `INS Dd.D[0],xn`, so this needs changed.

There is no intrinsic that lists FMOV as AArch64 Instruction. I don't mind changing it, though

CarolineConcatto: Ok, so there is no reason for this to be available in v7. Ok, A32/A64 is ok

CarolineConcatto: Martin I agree with George and  I am in favour to  update this to what the compilers are lowering to, which is FMOV

MartinWehking: ah, sorry yes, I updated this in the latest commit:
```
| <code>poly64x1_t <a href="https://developer.arm.com/architectures/instruction-sets/intrinsics/vmov_n_p64" target="_blank">vmov_n_p64</a>(poly64_t value)</code>                                                                                        | `value -> rn`                        | `FMOV Dd,rn`            | `Vd.1D -> result`  | `A32/A64`                 |
```

The other INS to FMOV changes are here:
https://github.com/ARM-software/acle/pull/445

CarolineConcatto (approved):
LGTM!

Kmeakin:
I don't see any reason not to make all the vreinterpret intrinsics available in all contexts, since they are a no-op at runtime. But if we're already gating some of them on AArch64, maybe better to leave as is, so LGTM

neon_intrinsics/advsimd.md
rockdreamer: I'm not sure about A32/A64 on many of these. ISTM only vtst_p16, vtstq_p16 are available on v7/A32/A64 and the rest are A64 only, or I haven't found a combination of switches that enables these. I'll send you an example.

MartinWehking: For the A64 only ones in LLVM/GCC: We will stick for the A32/A64 for now. This has been resolved internally.

For vtst_p16, vtstq_p16: If both are available on v7 in both compilers, I don't mind changing it

MartinWehking: Correction: I listed v7 for both of them already, so this doesn't need further changes.

rockdreamer: Yep, works for me

rockdreamer (approved):
LGTM

PR #441, opened 2026-07-08
[FMV] Add support-level table and remove __ARM_ACLE_VERSION

CarolineConcatto:
Remove __ARM_ACLE_VERSION, change __ARM_ACLE to represent when the header is available, and update __FUNCTION_MULTI_VERSIONING_SUPPORT_LEVEL so that it refers to a new Function Multi Versioning feature availability table.

For backwards compatibility with previous features, the table lists the support level at which each current FMV target version name became available. It also defines how future FMV feature sets allocate new support levels.

---
name: Pull request
about: Technical issues, document format problems, bugs in scripts or feature proposal.

---

**Thank you for submitting a pull request!**

If this PR is about a bugfix:

Please use the bugfix label and make sure to go through the checklist below.

If this PR is about a proposal:

We are looking forward to evaluate your proposal, and if possible to
make it part of the Arm C Language Extension (ACLE) specifications.

We would like to encourage you reading through the [contribution
guidelines](https://github.com/ARM-software/acle/blob/main/CONTRIBUTING.md), in particular the section on [submitting
a proposal](https://github.com/ARM-software/acle/blob/main/CONTRIBUTING.md#proposals-for-new-content).

Please use the proposal label.

As for any pull request, please make sure to go through the below
checklist.

Checklist: (mark with ``X`` those which apply)

* [ ] If an issue reporting the bug exists, I have mentioned it in the
      PR (do not bother creating the issue if all you want to do is
      fixing the bug yourself).
* [ ] I have added/updated the `SPDX-FileCopyrightText` lines on top
      of any file I have edited. Format is `SPDX-FileCopyrightText:
      Copyright {year} {entity or name} <{contact informations}>`
      (Please update existing copyright lines if applicable. You can
      specify year ranges with hyphen , as in `2017-2019`, and use
      commas to separate gaps, as in `2018-2020, 2022`).
* [ ] I have updated the `Copyright` section of the sources of the
      specification I have edited (this will show up in the text
      rendered in the PDF and other output format supported). The
      format is the same described in the previous item.
* [ ] I have run the CI scripts (if applicable, as they might be
      tricky to set up on non-*nix machines). The sequence can be
      found in the [contribution
      guidelines](https://github.com/ARM-software/acle/blob/main/CONTRIBUTING.md#continuous-integration). Don't
      worry if you cannot run these scripts on your machine, your
      patch will be automatically checked in the Actions of the pull
      request.
* [ ] I have added an item that describes the changes I have
      introduced in this PR in the section **Changes for next
[...]

PR #444, opened 2026-07-16
[FMV] Add support for more features

Lukacma:
The goal of this patch is to propose new features for function multi versioning. This should allow users to use FMV with newer features.

This patch also fixes a bug in register name for SME2 feature register.


mgabka:
I think it would make sense to split the "bug fix" part into a separate patch.

main/acle.md
```diff
@@ -3177,6 +3193,21 @@ The following table lists the feature dependencies for AArch64.
...
+   | ssve-fp8dot4     | sme2, fp8         |
+   | ssve-fp8fma      | sme2, fp8         |
+   | ssve-bitperm     | sme2              |
+   | lut              | simd              |
+   | faminmax         | simd              |
+   | sme-lutv2        | sme2              |
+   | sme2p1           | sme2              |
+   | sme2p2           | sme2p1            |
+   | sve2p1           | sve2              |
+   | sve2p2           | sve2p1            |
```
AlfieRichardsArm, line 3220: Not a bug, but I do wonder 

As sve2p2 requires armv9.5 is it worth (for FMV) implying everything thing that armv9.5 implies? That is not something we do for architecture options but it feels like there may be more to loose with FMV as it is more likely to be enabled via `target_version (sve2p2)` as opposed to `-march-armv9.5-a+sve2p2` or `-mcpu`?

AlfieRichardsArm (commented):
LGTM but I think this ought to wait for #441 so this can get its own FMV support version tick

main/acle.md
Lukacma: My main concern with this is that we can have situations where features added as optional from certain version will be moved and added as optional from earlier architecture versions in later releases.

Lukacma:
I have updated the patch with more features. If you guys could take another look that would be great!

mgabka:
Could you please rebase this patch? I think you also need to propagate the "The following table lists the architectures feature mapping for AArch64." with new `__FUNCTION_MULTI_VERSIONING_SUPPORT_LEVEL` value/values

Lukacma:
Thanks Maciej. Done. I have added it as a september version.

main/acle.md
```diff
@@ -3214,6 +3264,28 @@ The following table lists the feature dependencies for AArch64.
...
+   | sme-f16f16       | sme2              |
+   | d128             | lse               |
```
CarolineConcatto: Should this be lse128?
At least that is what I see in the compiler:
```
def FeatureD128 : ExtensionWithMArch<"d128", "D128",
  "FEAT_D128, FEAT_LVA3, FEAT_SYSREG128, FEAT_SYSINSTR128",
  "Enable Armv9.4-A 128-bit Page Table Descriptors, System Registers "
  "and instructions",
  [FeatureLSE128]>
```

main/acle.md
```diff
@@ -3214,6 +3264,28 @@ The following table lists the feature dependencies for AArch64.
...
    | sme-i16i64       | sme               |
    | sme2             | sme               |
+   | fp8              | simd              |
+   | f8f32mm          | simd, fp8         |
+   | fp8dot4          | fp8               |
+   | fp8fma           | fp8               |
+   | sme-f8f32        | sme2, fp8         |
+   | ssve-fp8dot4     | sme2, fp8         |
+   | ssve-fp8fma      | sme2, fp8         |
+   | ssve-bitperm     | sme2              |
```
CarolineConcatto, line 3288: Should this be sme2,sve-bitperm

```
def FeatureSSVE_BitPerm : ExtensionWithMArch<"ssve-bitperm", "SSVE_BitPerm", "FEAT_SSVE_BitPerm",
  "Enable Armv9.6-A SVE BitPerm support in streaming SVE mode", [FeatureSME2, FeatureSVEBitPerm]>;
```

main/acle.md
```diff
@@ -3214,6 +3264,28 @@ The following table lists the feature dependencies for AArch64.
    | sme-f64f64       | sme               |
    | sme-i16i64       | sme               |
    | sme2             | sme               |
+   | fp8              | simd              |
+   | f8f32mm          | simd, fp8         |
+   | fp8dot4          | fp8               |
+   | fp8fma           | fp8               |
+   | sme-f8f32        | sme2, fp8         |
+   | ssve-fp8dot4     | sme2, fp8         |
```
CarolineConcatto, line 3285: Why do new need the ssve-<smth> features? I would imagine we could still have them generated with +sme2 and the sve feature. Am I missing something?

main/acle.md
```diff
@@ -3102,6 +3102,31 @@ The following table lists the architectures feature mapping for AArch64.
...
+   | `FEAT_FAMINMAX`          | faminmax      | ```ID_AA64ISAR3_EL1.FAMINMAX >= 0b0001``` |
+   | `FEAT_SME_LUTv2`         | sme-lutv2     | ```ID_AA64SMFR0_EL1.LUTv2 == 0b1```       |
+   | `FEAT_SME2p1`            | sme2p1        | ```ID_AA64PFR1_EL1.SME >= 0b0001 && ID_AA64SMFR0_EL1.SMEver >= 0b0010```|
+   | `FEAT_SME2p2`            | sme2p2        | ```ID_AA64PFR1_EL1.SME >= 0b0001 && ID_AA64SMFR0_EL1.SMEver >= 0b0011```|
+   | `FEAT_SVE2p1`            | sve2p1        | ```ID_AA64PFR0_EL1.SVE >= 0b0001 && ID_AA64ZFR0_EL1.SVEver >= 0b0010```|
+   | `FEAT_SVE2p2`            | sve2p2        | ```ID_AA64PFR0_EL1.SVE >= 0b0001 && ID_AA64ZFR0_EL1.SVEver >= 0b0011```|
+   | `FEAT_SME_F16F16`        | sme-f16f16    | ```ID_AA64SMFR0_EL1.F16F16 == 0b1```      |
+   | `FEAT_GCS`               | gcs           | ```ID_AA64PFR1_EL1.GCS >= 0b0001```       |
+   | `FEAT_D128`              | d128          | ```ID_AA64MMFR3_EL1.D128 >= 0b0001```     |
+   | `FEAT_FP8DOT2`           | fp8dot2       | ```ID_AA64FPFR0_EL1.F8DP2 == 0b1```       |
```
CarolineConcatto: Do you mind moving this close to FEAT_FP8DOT4

main/acle.md
```diff
@@ -3102,6 +3102,31 @@ The following table lists the architectures feature mapping for AArch64.
    | `FEAT_SME2`              | sme2          | ```ID_AA64PFR1_EL1.SMEver >= 0b0001```    |
    | `FEAT_MOPS`              | mops          | ```ID_AA64ISAR2_EL1.MOPS >= 0b0001```     |
    | `FEAT_CSSC`              | cssc          | ```ID_AA64ISAR2_EL1.CSSC >= 0b0001```     |
+   | `FEAT_FP8`               | fp8           | ```ID_AA64FPFR0_EL1.F8CVT == 0b1```       |
```
CarolineConcatto, line 3107: I've noticed that for FP8 the reference informs other system registers, should this be checked as well?
The following fields identify the presence of FEAT_FP8:
• ID_AA64FPFR0_EL1.F8CVT. • ID_AA64FPFR0_EL1.F8E4M3. • ID_AA64FPFR0_EL1.F8E5M2.

Lukacma: That is not necessary imo. Reading the spec for the F8CVT:

> [FEAT_FP8](https://support.arm.com/documentation/ddi0487/maa/-Part-A-Arm-Architecture-Introduction-and-Overview/-Chapter-A2-A-profile-Architecture-Extensions/-A2-3-Armv9-A-architecture-extensions/-A2-3-6-The-Armv9-5-architecture-extension?lang=en#feat_feat_fp8) implements the functionality identified by the value 1.

To me it means that when the value is 1 FEAT_FP8 is implemented

main/acle.md
Lukacma: Sure we could them generated like that, but that is only true for cores which have both non-ssve and ssve feature. If core supports only ssve variant that function version wouldn't be emitted.

main/acle.md
Lukacma: We don't have sve-bitperm feature in FMV. We have sve2-bitperm, which enables both sve2 and sve-bitperm feature and is not a dependency of ssve-bitperm.

main/acle.md
Lukacma: We don't have lse128 feature in FMV as we internally agreed to add features which introduce new builtins and intrinsics. So I went for indirect dependency here as lse128 requires lse.

main/acle.md
Lukacma: Done

main/acle.md
CarolineConcatto: I would also move this close to the f8fdot4

main/acle.md
```diff
@@ -3272,6 +3278,9 @@ The following table lists the feature dependencies for AArch64.
    | ssve-fp8dot4     | sme2, fp8         |
    | ssve-fp8fma      | sme2, fp8         |
    | ssve-bitperm     | sme2              |
+   | ssve-fp8dot2     | sme2, fp8         |
```
CarolineConcatto: Can we group these together too ssve-fp8dot4 and ssve-fp8dot2

main/acle.md
```diff
@@ -3102,6 +3102,34 @@ The following table lists the architectures feature mapping for AArch64.
    | `FEAT_SME2`              | sme2          | ```ID_AA64PFR1_EL1.SMEver >= 0b0001```    |
    | `FEAT_MOPS`              | mops          | ```ID_AA64ISAR2_EL1.MOPS >= 0b0001```     |
    | `FEAT_CSSC`              | cssc          | ```ID_AA64ISAR2_EL1.CSSC >= 0b0001```     |
+   | `FEAT_FP8`               | fp8           | ```ID_AA64FPFR0_EL1.F8CVT == 0b1```       |
+   | `FEAT_F8F32MM`           | f8f32mm       | ```ID_AA64FPFR0_EL1.F8MM8 == 0b1```       |
```
CarolineConcatto, line 3108: Being a bit pendantic, I would order this fetaures in the following order:
scalar/neon/sve/ssve/sme.
Unless this order has some reason to be the way it is.

Lukacma: The order determines priority in function selection, so the feature dependencies need to be observed. But I will  try to maintain this ordering, while preserving dependencies.

main/acle.md
Lukacma: done

main/acle.md
Lukacma: done

main/acle.md
Lukacma: Reordered

main/acle.md
mgabka: > We don't have lse128 feature in FMV as we internally agreed to add features which introduce new builtins and intrinsic

FEAT_LSE128 adds new instructions, I can imagine user may want to to enable versioning based on the presence of this feature as it affects codegen, you can see it for example here https://godbolt.org/z/nPzarGzzP

PR #452, opened 2026-08-18
Correct a few Neon entries

MartinWehking:
This includes:
- mfloat8 is limited to A64 always
- Replace some lower case letters with upper case ones where appropriate
- Fix some arg mappings (e.g. "arrangement specifiers")


rockdreamer (commented):
One nit but LGTM other than that

tools/intrinsic_db/advsimd.csv
```diff
@@ -2075,7 +2075,7 @@ float32x2_t vget_low_f32(float32x4_t a)	a -> Vn.4S	DUP Vd.1D,Vn.D[0]	Vd.2S -> re
 poly8x8_t vget_low_p8(poly8x16_t a)	a -> Vn.16B	DUP Vd.1D,Vn.D[0]	Vd.8B -> result	v7/A32/A64
 poly16x4_t vget_low_p16(poly16x8_t a)	a -> Vn.8H	DUP Vd.1D,Vn.D[0]	Vd.4H -> result	v7/A32/A64
 float64x1_t vget_low_f64(float64x2_t a)	a -> Vn.2D	DUP Vd.1D,Vn.D[0]	Vd.1D -> result	A64
-mfloat8x8_t vget_low_mf8(mfloat8x16_t a)	a -> Vn.16B	DUP Vd.1D,Vn.D[0]	Vd.8B -> result	A64
+mfloat8x8_t vget_low_mf8(mfloat8x16_t a)	a -> Vd.16B	NOP	Vd.8B -> result	A64
```
rockdreamer: This was likely a MOV?

MartinWehking: No, I deliberately picked NOP here, because neither clang/gcc emit a mov/dup for this intrinsic (regardless of the optimization level) There are other intrinsics also lowering to a NOP.

Is this semantically ok?

rockdreamer: Yep, that's correct. Thanks

rockdreamer (approved):

rockdreamer (approved):

PR #457, opened 2026-08-25
CI: add link check exceptions for frequent 429s

rgwott:
Currently, the markdown-link-check workflow frequently fails not because some link is actually broken, but because we often get 429s (too many requests) on two domains (gnu and sourceware). In this patch, we purposely ignore those 429s and let the workflow pass in this case.


mgabka (approved):

rockdreamer (approved):
LGTM, it's sad that some antisocial people cause problems for all users of this site :( .

PR #459, opened 2026-08-27
CI: add push actions; update action versions

rgwott:
- Trigger tag/deploy workflows automatically on push/PR merge.
- Update action versions to newer ones with Node 24.
- Delete Latest release before creating a new Latest.
- Modify text of Latest release (currently the message of the latest commit).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,9 @@ name: CI
 
 on:
   pull_request:
+  push:
+    branches:
+      - main
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -5,17 +5,20 @@ name: CI
...
+      - main
   workflow_dispatch:
 
 permissions:
   contents: read
 
 jobs:
   build-pdfs:
-    if: github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'
+    if: github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch' || (github.event_name == 'push' && github.ref == 'refs/heads/main')
```
CarolineConcatto, line 18: Why the pdf and the mark downs dont need to check :
` github.repository == 'ARM-software/acle'`

CarolineConcatto (commented):
Why are ups updating the version?

.github/workflows/ci.yml
rgwott: Those can be executed in any case. The ones that deploy tags and the website have this safeguard so that forks don't execute them (even though they should not succeed).

rgwott:
@CarolineConcatto We are updating the versions because the images used until now are deprecated, using Node 20 (which has security flaws). The recommended approach is to use the newer versions.

CarolineConcatto (approved):
LGTM

PR #460, opened 2026-08-27
CI: cosmetic changes (short hash)

rgwott:
Standardize all commit hashes to 7-character short form.


README.md
```diff
@@ -56,18 +56,13 @@ an email to arm.acle@arm.com.
...
-
-[Arm Neon Intrinsics Reference](https://arm-software.github.io/acle/neon_intrinsics/)
-
-[Arm®v8-M Security Extensions - Requirements on Development Tools](https://arm-software.github.io/acle/cmse/)
-
-The PDF versions can be found in the latest GitHub
-[release](https://github.com/ARM-software/acle/releases/latest).
+HTML version                                                                                                  | PDF version
+---                                                                                                           | ---
+[Arm C Language Extensions](https://arm-software.github.io/acle/main/)                                        | [download](https://github.com/ARM-software/acle/releases/latest)
```
mgabka: All the links in the PDF version column point to the same GitHub release page rather than to the corresponding PDF file. Calling them download links therefore seems a little misleading. Could these link directly to the corresponding PDF?

mgabka (commented):
Thanks Rodolfo for the changes, they all look like useful improvements

rockdreamer (approved):

README.md
rgwott: That should not be possible unless we give the pdfs generic filenames without the commit hashes.
For the record, given how legacy releases also have this problem (e.g. files called acle-2026Q1.pdf and such), this is how it already was.

mgabka: would it be possible to handle it with some sort of generics symlinks which  destinations would be changed every time we generate new releases? Or generating new URL every time we create new pdf?

rgwott: This should be possible when building the webpages. I will look into this.

rgwott: There is no simple solution to generate the links dynamically for the webpage but leaving the README static on github, so let's not restore the table after all and just fix the short-form hashes in this patch.

mgabka (approved):

